PR #14, opened 2026-02-11
Feat: Add multi-window chunking

richzw:
# Multi-Window Chunking

## Summary

Add an opt-in multi-window chunking strategy inspired by [AI21 Labs' multi-scale retrieval](https://github.com/AI21Labs/multi-window-chunk-size). Instead of indexing each document with a single chunk size, we generate an additional set of **wide-window chunks** by merging adjacent heading sections with a sliding window, then store both windows in the same Milvus collection. The existing hybrid search (dense + BM25 + RRF) naturally fuses results from different granularities without any extra retrieval passes.

This is a **lightweight variant** of AI21's approach — it avoids the overhead of maintaining multiple collections or multiplying storage by N, while still capturing cross-heading-boundary context that the current heading-based chunking misses.

## Motivation

The current chunker splits on markdown headings. This works well for self-contained sections, but loses context that spans heading boundaries. Example:

```markdown
## Redis
We chose Redis for caching because of low latency.

## Configuration
Set `REDIS_PORT=6379` in your `.env` file.
```

A query like **"Redis configuration"** won't get a strong match from either chunk alone, because the relevant information is split across two sections. A wide chunk merging both sections would capture this cross-boundary context.

### Core Idea

```
Standard window (existing):  [Title] [A] [B] [C]        — heading-based splits
Wide window (new):           [Title+A] [A+B] [B+C]      — sliding window over adjacent sections
```

Both window types share the **same Milvus collection**, distinguished by a `window` metadata field. The existing RRF reranker naturally fuses results from both windows.

### Architecture

```
markdown file
     │
     ▼
_extract_sections()           ← shared section extraction
     │
     ├───────────────────────────┐
     ▼                           ▼
Standard window              Wide window
_sections_to_chunks()        _build_wide_chunks()
window="standard"            window="wide"
     │                           │
     └─────────┬─────────────────┘
               ▼
         merged chunks list
               │
               ▼
     compute_chunk_id()       ← unchanged formula, different content → different IDs
               │
               ▼
      dedup → embed → upsert into single collection
               │
               ▼
    hybrid search (dense + BM25 + RRF)   ← both windows compete on merit
               │
               ▼
    _dedupe_overlapping()     ← remove lower-scored results covered by higher-scored ones
               │
               ▼
         top_k results
```

### Config file (recommended)

```toml
[chunking]
multi_window = true
wide_window_size = 2          # merge 2 adjacent sections
max_wide_chunk_size = 4000    # skip merges exceeding this
```

### CLI flag

```bash
memsearch index ./docs --multi-window
memsearch search "Redis configuration" --multi-window
```

### Environment variable

```bash
export MEMSEARCH_CHUNKING_MULTI_WINDOW=true
```

--------

## Note

This is one rough idea, Glad to discuss with you for more details.

### DIFF
--- a/docs/claude-plugin.md
+++ b/docs/claude-plugin.md
@@ -166,46 +166,212 @@ On every user prompt, the `UserPromptSubmit` hook injects the top-3 semantic sea
 
 This happens transparently -- no action from Claude or the user is required.
 
-**Example injection:**
+**Example injection** (this is what Claude sees before processing each message):
 
 ```
 ## Relevant Memories
-- [memory/2026-02-08.md:Session 14:30]  Implemented caching system with Redis L1
-  and in-process LRU L2. Fixed N+1 query issue in order-service using selectinload...
-  `chunk_hash: a1b2c3d4e5f6`
+- [.memsearch/memory/2026-02-12.md:04:16]  使用远程 Milvus Server 测试
+  `claude -p` 模式，索引了 4 个 chunks，search 正常返回相关结果。确认
+  `claude -p` 模式不触发任何 hooks...
+  `chunk_hash: 47b5475122b992b6`
+- [.memsearch/memory/2026-02-11.md:11:02]  修复 watch 启动时不索引已有文件的
+  问题。ccplugin stop.sh 增加显式 memsearch index 调用...
+  `chunk_hash: 31cbaf74856ad1ed`
 ```
+
+The preview is enough for Claude to answer most follow-up questions. But when it needs the full picture, it moves to L2.
 
 ### L2: On-Demand Expand
 
-When an L1 preview is not enough, Claude can run the `memsearch expand` command to retrieve the full markdown section surrounding a chunk:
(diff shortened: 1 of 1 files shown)


richzw:
> @richzw thanks, currently, our chunking has a slight issue, and we are considering optimizing and restructuring it. The multi-granularity chunking you mentioned might be a solution, but it will significantly increase storage costs. We are still evaluating which specific approach to adopt.

@zc277584121 , Thank you for your reply.

You’re right — multi-granularity chunking increases storage cost. For complex query scenarios, if we want higher retrieval accuracy, we may have to accept higher storage cost; it’s a trade-off and needs balancing.

I’m thinking we could use a progressive strategy: when storage is small and the directory structure is simple, use naive chunking; when storage grows and there are more directories (i.e., the corpus becomes more heterogeneous), switch to multi-granularity chunking. The decision could be based on a threshold, or we could have an AI model make the call. Just a rough, not-fully-baked idea.  :smile:

PR #70, opened 2026-02-19
fix: batch embedding requests to respect provider API limits

### LINKED ISSUE #72: [BUG] Google embedding provider fails with >100 chunks: BatchEmbedContentsRequest limit
## Bug

`memsearch index` crashes with `ClientError: 400 INVALID_ARGUMENT` when a markdown file produces more than 100 chunks while using the Google embedding provider (`gemini-embedding-001`).

## Environment

- **memsearch**: 0.1.11
- **Python**: 3.13.12 (pyenv)
- **OS**: macOS 15 (Apple Silicon)
- **Provider**: `google` / `gemini-embedding-001`
- **Config**: default `max_chunk_size=1500`, `overlap_lines=2`

## Steps to Reproduce

1. Configure Google as embedding provider:
   ```toml
   [embedding]
   provider = "google"
   model = "gemini-embedding-001"
   ```
2. Have a markdown file larger than ~75KB (mine was 96KB producing ~150 chunks)
3. Run `memsearch index .memsearch/memory/`

## Error

```
google.genai.errors.ClientError: 400 INVALID_ARGUMENT.
{'error': {'code': 400, 'message': '* BatchEmbedContentsRequest.requests: at most 100 requests can be in one batch\n', 'status': 'INVALID_ARGUMENT'}}
```

## Root Cause

[`_embed_and_store()`](https://github.com/zilliztech/memsearch/blob/main/src/memsearch/core.py#L140-L146) in `core.py` sends **all** chunks to `self._embedder.embed()` in a single call with no batching.

Google's Gemini Embedding API has a hard limit of **100 items per batch request**. Other providers also have limits (Voyage: 128, OpenAI: 2048), so this will eventually fail for any provider with large enough files.

## Suggested Fix

Add batching in [`_embed_and_store()`](https://github.com/zilliztech/memsearch/blob/main/src/memsearch/core.py#L140-L146), ideally with a provider-configurable `max_batch_size` property on each [embedding provider class](https://github.com/zilliztech/memsearch/tree/main/src/memsearch/embeddings):

| Provider | Batch Limit |
|----------|-------------|
| Google Gemini | 100 |
| Voyage | 128 |
| OpenAI | 2048 |
| Ollama / Local | unlimited |

svidskiy:
## Summary

- Adds `max_batch_size` property to `EmbeddingProvider` Protocol and all 5 provider implementations (Google: 100, Voyage: 128, OpenAI: 2048, Ollama: 512, Local: 512)
- Modifies `_embed_and_store()` in `core.py` to batch embedding requests according to each provider's limit instead of sending all chunks in a single API call
- Adds tests for batching logic in `tests/test_embed_batching.py`

Fixes #72 

## Test plan

- [x] `uv run python -m pytest tests/test_embed_batching.py` — 4 tests pass
- [x] Manual test with Google Gemini provider on a file producing >100 chunks

### DIFF
--- a/src/memsearch/core.py
+++ b/src/memsearch/core.py
@@ -143,7 +143,13 @@ class MemSearch:
 
         model = self._embedder.model_name
         contents = [c.content for c in chunks]
-        embeddings = await self._embedder.embed(contents)
+
+        # Batch embedding requests to respect provider-specific API limits
+        batch_size = self._embedder.max_batch_size
+        embeddings: list[list[float]] = []
+        for i in range(0, len(contents), batch_size):
+            batch = contents[i:i + batch_size]
+            embeddings.extend(await self._embedder.embed(batch))
 
         records: list[dict[str, Any]] = []
         for i, chunk in enumerate(chunks):
--- a/src/memsearch/embeddings/__init__.py
+++ b/src/memsearch/embeddings/__init__.py
@@ -15,6 +15,9 @@ class EmbeddingProvider(Protocol):
     @property
     def dimension(self) -> int: ...
 
+    @property
+    def max_batch_size(self) -> int: ...
+
     async def embed(self, texts: list[str]) -> list[list[float]]: ...
 
 
--- a/src/memsearch/embeddings/google.py
+++ b/src/memsearch/embeddings/google.py
@@ -35,6 +35,10 @@ class GoogleEmbedding:
     def dimension(self) -> int:
         return self._dimension
 
+    @property
+    def max_batch_size(self) -> int:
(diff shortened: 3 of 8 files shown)


zc277584121:
Thank you @svidskiy for identifying this issue and providing a solid fix with tests\! We've merged an alternative approach in #83 that keeps batching as an internal implementation detail of each provider (without modifying the `EmbeddingProvider` protocol) and makes batch size user-configurable via TOML config / CLI flag. Your work on `test_embed_batching.py` was especially helpful as a reference. Closing in favor of #83.

PR #74, opened 2026-02-20
fix: deduplicate stop-hook summaries with deferred summarization

svidskiy:
## Summary

The `Stop` hook fires after **every assistant turn**, not just at session end. This causes N near-identical summaries to be appended to the daily `.md` file for a single session — polluting both the memory files and the Milvus index with duplicates.

**Example (before):** a session with 3 Claude responses produces 3 almost identical entries:

```markdown
## Session 13:35

### 13:35
- Removed symlinks, copied skill directories, deleted .agents/

### 13:36
- Removed symlinks, copied skill directories, deleted .agents/   ← duplicate
- Created .memsearch.toml

### 13:36
- Removed symlinks, copied skill directories, deleted .agents/   ← duplicate
- Created .memsearch.toml
```

### Changes

This PR introduces **deferred summarization** — the Stop hook becomes a lightweight state recorder, and the actual LLM summarization happens exactly once at session end:

- **`stop.sh`** — stripped down from 120 to 45 lines. Now only saves the transcript path to a pending state file (`.memsearch/.pending/<session_id>`). Each turn overwrites the same file (last-writer-wins). No parsing, no LLM call — instant and free.

- **`common.sh`** — new functions:
  - `_summarize_pending()` — reads transcript path from state file, parses transcript, calls `claude -p` to summarize, appends to daily `.md`
  - `flush_pending()` — iterates pending state files, summarizes each, re-indexes
  - `_check_embedding_key()` — extracted API key validation as a reusable helper

- **`session-end.sh`** — calls `flush_pending()` before `stop_watch` (primary flush path)

- **`session-start.sh`** — calls `flush_pending()` as a safety net for crashed sessions where `SessionEnd` didn't fire

### Benefits

- **No duplicates** — exactly 1 summary per session, by design (not heuristics)
- **N→1 LLM calls** — a session with 10 turns now makes 1 Haiku call instead of 10
- **Faster Stop hook** — writing 2 lines to a file vs. parsing + LLM roundtrip on every turn
- **Crash-safe** — `SessionStart` picks up any unflushed state from previous sessions

Also adds `MEMSEARCH_SUMMARY_MODEL` env var (default: `haiku`) to configure the summarization model (`haiku` | `sonnet` | `opus`).

---

> **Note:** This is a proposal. If there's interest in improving the ccplugin hook logic for Claude Code, I'd be happy to iterate on this approach and develop it further. In my experience this is the most efficient solution for the duplication problem, but I'm open to feedback.

## Test plan

- [x] Run a multi-turn session and verify only 1 summary appears in the daily `.md` after session end
- [x] Kill a session mid-conversation, start a new one, and verify `SessionStart` flushes the orphaned pending file
- [x] Set `MEMSEARCH_SUMMARY_MODEL=sonnet` and verify the summary uses Sonnet
- [x] Verify `.memsearch/.pending/` is cleaned up after flush

### DIFF
--- a/ccplugin/hooks/common.sh
+++ b/ccplugin/hooks/common.sh
@@ -1,26 +1,31 @@
 #!/usr/bin/env bash
-# Shared setup for memsearch command hooks.
-# Sourced by all hook scripts — not executed directly.
+# common.sh — shared utilities for all memsearch hooks.
+# Sourced (not executed) by every hook script.
 
 set -euo pipefail
 
-# Read stdin JSON into $INPUT
+# shellcheck disable=SC2034  # INPUT is read by scripts that source this file.
 INPUT="$(cat)"
 
-# Ensure common user bin paths are in PATH (hooks may run in a minimal env)
+# Hooks may run in a minimal environment; ensure common bin paths are available.
 for p in "$HOME/.local/bin" "$HOME/.cargo/bin" "$HOME/bin" "/usr/local/bin"; do
   [[ -d "$p" ]] && [[ ":$PATH:" != *":$p:"* ]] && export PATH="$p:$PATH"
 done
 
-# Hooks directory (where this file lives)
-HOOKS_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+# ---------------------------------------------------------------------------
+# Paths
+# ---------------------------------------------------------------------------
 
-# Memory directory and memsearch state directory are project-scoped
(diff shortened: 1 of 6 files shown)


svidskiy:
If you're interested, I can also update the docs (`docs/claude-plugin.md`) to reflect the new deferred summarization flow — the current Stop hook section, lifecycle diagram, and troubleshooting guide still describe the old per-turn behavior. Happy to include that in this PR or as a follow-up.

PR #120, opened 2026-03-05
feat: add /commit skill for conventional commits

svidskiy:
## Summary

- Add a project-level Claude Code skill at `.claude/skills/commit/SKILL.md`
- Invoked with `/commit` — creates conventional commits following the project's prefix conventions
- `disable-model-invocation: true` — Claude cannot trigger it automatically, only the user can
- Supports all prefixes from CONTRIBUTING.md: `feat`, `fix`, `docs`, `ci`, `chore`, `refactor`, `test`, `style`, `perf`
- Compatible with release-drafter autolabeler regex (handles `prefix:`, `prefix(scope):`, `prefix!:`)

### How it works

The skill instructs Claude to:
1. Check `git status` and `git diff` to understand changes
2. Pick the correct conventional commit prefix
3. Stage files by name (never `git add -A`)
4. Write a commit message with optional scope: `prefix(scope): description`
5. Never push, amend, or commit secrets

### Usage

```
/commit
/commit fix the scanner edge case
```

## Test plan

- [x] Skill file follows [Claude Code skills docs](https://code.claude.com/docs/en/skills): YAML frontmatter, `disable-model-invocation: true`, `allowed-tools`
- [x] All prefixes match release-drafter autolabeler config in #119
- [x] Skill does not auto-trigger — user-only invocation

### DIFF
--- a/.claude/skills/commit/SKILL.md
+++ b/.claude/skills/commit/SKILL.md
@@ -1,0 +1,45 @@
+---
+name: commit
+description: Create a conventional commit following project conventions
+argument-hint: "[optional message override]"
+disable-model-invocation: true
+allowed-tools: Bash(git *)
+---
+
+Create a commit following this project's conventions.
+
+## Steps
+
+1. Run `git status` to see changed files (never use `-uall`).
+2. Run `git diff` (staged + unstaged) to understand what changed.
+3. Run `git log --oneline -5` to see recent commit style.
+4. Choose the correct conventional commit prefix based on the changes:
+
+   | Prefix     | When to use                                    |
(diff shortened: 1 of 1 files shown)


zc277584121 (commented):
Nice skill — the conventional commit table and rules are well written.

One consideration: this adds a `/commit` skill to the project-level `.claude/` directory, so every Claude Code user working in this repo will get it. That's probably fine for this project, but just want to make sure that's intentional.

LGTM on content. Ruff CI is failing — likely needs a rebase onto latest main (after #116 formatting merge).

PR #208, opened 2026-03-20
fix: prevent single file failure from blocking entire index

samlevan:
## Summary

- **Error isolation**: Wrap `_index_file()` in try/except during bulk `index()` so one file failing doesn't prevent subsequent files from being indexed. Failed files are logged and skipped.
- **Token-aware batching**: Add `max_tokens` budget to `batched_embed()` that splits batches when estimated tokens would exceed the provider's API limit (default 250K, conservative vs OpenAI's 300K cap). Previously, batching only checked item count (default 2,048), so a file with 1,000+ chunks could exceed the token limit in a single API call.

### The problem

If any markdown file produces chunks whose total tokens exceed OpenAI's 300K-token-per-request embedding limit, `_index_file()` raises a `400 BadRequestError`. Since there's no try/except in the `index()` loop, this kills the entire indexing process. All files sorted after the failing one are never indexed.

This is made worse by the error being silently swallowed in the Claude Code plugin hooks (`2>/dev/null`, `|| true`, `&>/dev/null`), so users have no indication their index is incomplete.

### Real-world scenario

A single large memory file (1.7MB, ~1,000 chunks, ~425K tokens) crashed indexing on every run. Because files are processed alphabetically and this was the 2nd file, only 1 of 27 memory files was ever indexed. Search returned irrelevant results because 96% of the corpus was missing from the index.

## Test plan

- [x] New test: `test_batched_embed_splits_on_token_limit` - verifies batches split when tokens exceed limit even if item count is under `batch_size`
- [x] New test: `test_batched_embed_respects_both_limits` - verifies both item and token limits are enforced
- [x] New test: `test_estimate_tokens` - token estimation from character length
- [x] New test: `test_split_into_batches_by_tokens` and `test_split_into_batches_by_items` - batch splitting logic
- [x] New test: `test_index_continues_after_file_failure` - verifies index() continues past a failing file
- [x] All 14 batching tests pass, full suite 77/78 pass (1 pre-existing flaky OpenAI determinism test)

### DIFF
--- a/src/memsearch/core.py
+++ b/src/memsearch/core.py
@@ -90,11 +90,16 @@ class MemSearch:
         """
         files = scan_paths(self._paths)
         total = 0
+        failed = 0
         active_sources: set[str] = set()
         for f in files:
             active_sources.add(str(f.path))
-            n = await self._index_file(f, force=force)
-            total += n
+            try:
+                n = await self._index_file(f, force=force)
+                total += n
+            except Exception:
+                failed += 1
+                logger.exception("Failed to index %s, skipping", f.path)
 
         # Clean up chunks for files that no longer exist
         indexed_sources = self._store.indexed_sources()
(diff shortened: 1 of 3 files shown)


zc277584121:
Thank you for identifying this issue and the detailed write-up! The error isolation approach (try/except in `index()`) was exactly right.

We adopted the first layer of your fix (error isolation in `core.py`) in #215, with a simpler approach for the batch-size issue: lowering OpenAI's `_DEFAULT_BATCH_SIZE` from 2048 to 256 instead of adding token estimation logic, to keep `batched_embed()` provider-agnostic.

Closing this PR since the fix has been merged via #215. Thanks again for the contribution!

PR #214, opened 2026-03-22
docs: clarify compact --source path behavior

haosenwang1018:
## Summary
- document that `compact --source` now accepts both relative and absolute paths
- explain that existing relative / `~` paths are normalized to the absolute indexed source path
- note that the CLI prints the resolved source path when no chunks match, making filter mismatches easier to debug

## Testing
- docs-only change


zc277584121:
The code changes in this PR were already merged via #213. The docs-only additions (README + cli.md) will be included in a follow-up commit. Thanks for the thorough documentation!

PR #289, opened 2026-04-06
fix: remove unreachable UUID prefix-match in transcript parser

j-sperling:
## Summary
- Remove dead `target_uuid.startswith(turn.uuid[:8])` condition in `find_turn_context()`
- All callers pass full UUIDs; the 8-char prefix match is unreachable
- Eliminates theoretical false-positive risk in long transcripts

## Test plan
- [x] Existing transcript tests still pass
- [x] Verified: all HTML comment anchors in memory files contain full UUIDs

### DIFF
--- a/plugins/claude-code/transcript.py
+++ b/plugins/claude-code/transcript.py
@@ -123,7 +123,7 @@ def find_turn_context(
     """
     target_idx = -1
     for i, turn in enumerate(turns):
-        if turn.uuid.startswith(target_uuid) or target_uuid.startswith(turn.uuid[:8]):
+        if turn.uuid.startswith(target_uuid):
             target_idx = i
             break
 


zc277584121 (changes requested):
The change makes matching stricter, which is generally good. However, I want to make sure the removed condition is truly unreachable before merging.

The removed branch (`target_uuid.startswith(turn.uuid[:8])`) handles the case where the target UUID is longer/more complete than the stored turn UUID. This could happen if:
- The chunk anchor stores a full UUID but the JSONL transcript has truncated UUIDs
- Or vice versa

Could you verify by checking a few real JSONL transcript files (`~/.claude/projects/*/sessions/*/transcript.jsonl`) — are the UUIDs there always full-length, or are there cases where they're truncated?

If turn UUIDs are always complete, then `turn.uuid.startswith(target_uuid)` alone is sufficient and this PR is correct. If there are truncated cases, we should keep a reverse match but with a longer prefix (e.g. 16 chars instead of 8) to reduce false positives.

PR #301, opened 2026-04-08
chore: switch from uv add to uv tool install for CLI availability

liziwl:
**Title:** chore: switch from `uv add` to `uv tool install` for CLI availability

**Description:**
This PR replaces `uv add` with `uv tool install` for the specified package.

**Reasoning:**
- `uv add` is designed for managing project-specific dependencies within a virtual environment (`pyproject.toml`). It does not make the CLI globally accessible.
- `uv tool install` installs the package in an isolated environment and exposes the executable globally (via PATH), allowing the CLI tool to be run directly from the terminal without activating a specific project environment.

This change ensures the tool is properly installed as a standalone global utility rather than a project dependency.

### DIFF
--- a/README.md
+++ b/README.md
@@ -322,16 +322,20 @@ User: "What did we discuss about batch size?"
 # pip
 pip install memsearch
 
-# or uv (recommended)
+# Install as a CLI tool (recommended for most users)
 uv tool install memsearch
+
+# Or add as a project dependency
+uv add memsearch
 ```
 
 <details>
 <summary><b>Optional embedding providers</b></summary>
 
 ```bash
-pip install "memsearch[onnx]"    # Local ONNX (recommended, no API key)
-# or: uv tool install "memsearch[onnx]"
+pip install "memsearch[onnx]"          # Local ONNX (recommended, no API key)
+# uv tool install "memsearch[onnx]"    # CLI with ONNX embedding
+# or: uv add "memsearch[onnx]"         # as project dependency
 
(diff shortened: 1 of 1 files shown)


zc277584121 (changes requested):
Thanks for the fix! `uv tool install` is indeed the right command for installing CLI tools globally. However, `uv add` is still correct for users who want to add memsearch as a project dependency (library usage).

Could you update to show both options instead of a simple replacement? Something like:

```bash
# Install as a CLI tool (recommended for most users)
uv tool install memsearch

# Or add as a project dependency
uv add memsearch
```

And similarly for the extras section:

```bash
uv tool install "memsearch[onnx]"    # CLI with ONNX embedding
# or: uv add "memsearch[onnx]"       # as project dependency
```

liziwl:
@zc277584121 Thanks for the review! Updated to show both installation options as suggested.

zc277584121:
Thanks @liziwl for spotting this! Your diagnosis was exactly right — `uv add` only installs into the current project's venv, so recommending it as the primary uv install was wrong for the common case where users just want the `memsearch` CLI available globally (which is what every plugin needs).

We took a slightly broader approach in #411 (just merged): the install block now splits into two explicit groups — **CLI tool** (`uv tool install` / `pipx install` / `pip install`) for the common case, and **project dependency** (`uv add`) for users who want to import `memsearch` from their own Python code. The optional-extras section got the same treatment.

Closing this as superseded — really appreciate you catching the issue.

PR #305, opened 2026-04-08
feat: add indexed memory listing

officialasishkumar:
## Summary
- add `list_memories()` to the store and core layers so indexed chunks can be enumerated deterministically
- add a new `memsearch list` command with optional source-prefix filtering, JSON output, and preview-friendly text formatting
- document the new listing workflow in the README, CLI docs, and Python API docs
- add pytest coverage for store ordering, the core API, CLI output, and CLI help

Closes #277.

## Validation
- `.venv/bin/ruff check src tests`
- `.venv/bin/ruff format --check src tests`
- `.venv/bin/python -m pytest`

### DIFF
--- a/README.md
+++ b/README.md
@@ -535,6 +535,7 @@ memsearch index ./memory/                          # index markdown files
 memsearch index ./memory/ ./notes/ --force         # re-embed everything
 memsearch search "Redis caching"                   # hybrid search (BM25 + vector)
 memsearch search "auth flow" --top-k 10 --json-output  # JSON for scripting
+memsearch list --limit 20                          # inspect indexed memories without guessing a query
 memsearch expand <chunk_hash>                      # show full section around a chunk
 ```
 
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -346,6 +346,39 @@ $ memsearch search "database migrations" --provider google
 
 ---
 
+## `memsearch list`
+
+List indexed memories without needing a search query first. Results are ordered by source path and line number so you can quickly inspect what is already in the memory index.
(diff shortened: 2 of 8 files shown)


zc277584121:
Thanks for the PR and the thoughtful implementation! After reviewing this alongside #320, #367, and the underlying use case in #277, we decided **not to add a dedicated `memsearch list` CLI command**. Rationale:

memsearch's design philosophy is "markdown is source of truth, Milvus is a derived index". The canonical way to see what's been memorized is the daily .md files themselves:

```bash
ls -t .memsearch/memory/ | head -10
grep -h "^## " .memsearch/memory/*.md | sort -u | tail -40
cat .memsearch/memory/<YYYY-MM-DD>.md
```

Adding a `list` command on top of this would:
- reinvent weaker `ls`/`grep`/`cat` (which compose into arbitrary pipelines)
- leak the "chunk" abstraction — users/agents should reason in terms of files and sessions, not Milvus chunks
- duplicate what the SessionStart cold-start injection is already supposed to do

What we did instead (PR #410, just merged): upgrade the cold-start injection on all 4 platform plugins from `tail -30` to a heading+bullet summary, so agents see the structure of recent memory files (which sessions existed, what topics came up). And we added a "When unsure what to search" section to all 4 `memory-recall` SKILL.md files pointing agents at `ls`/`grep`/`cat` when they can't form a concrete search query.

That addresses the underlying ask in #277 without expanding the CLI surface. Closing this PR — really appreciate the clean implementation and the test coverage, and thanks for bringing attention to this area.

PR #307, opened 2026-04-08
feat: add direct Claude Code memory skills

DarKWinGTM:
## Summary
- add direct Claude Code plugin skills for `memory-search`, `memory-expand`, `session-recall`, `memory-stats`, and `config-check`
- add a bounded `memory-router` orchestration/front-door skill so Claude can choose the correct memsearch retrieval path before broader fallback behavior
- tighten `session-recall` so bare session-id lookups still start with memsearch search/expand before bounded fallback
- let the Claude Code memory skills use a broader search/read tool pool while keeping `memsearch search` as the preferred first path and keeping fallback behavior bounded
- document how the direct skills, `memory-recall`, and `memory-router` fit together in the Claude Code plugin docs
- extend Claude Code skill-surface tests to cover the direct skills and router/session-recall contract

## Updated files
- `.claude-plugin/marketplace.json`
- `docs/platforms/claude-code/how-it-works.md`
- `docs/platforms/claude-code/index.md`
- `docs/platforms/claude-code/installation.md`
- `docs/platforms/claude-code/memory-recall.md`
- `docs/platforms/claude-code/memory-tools.md`
- `mkdocs.yml`
- `plugins/claude-code/README.md`
- `plugins/claude-code/skills/config-check/SKILL.md`
- `plugins/claude-code/skills/memory-expand/SKILL.md`
- `plugins/claude-code/skills/memory-router/SKILL.md`
- `plugins/claude-code/skills/memory-search/SKILL.md`
- `plugins/claude-code/skills/memory-stats/SKILL.md`
- `plugins/claude-code/skills/session-recall/SKILL.md`
- `tests/test_claude_code_skills.py`
- `uv.lock`

## Test plan
- [x] `uv run python -m pytest tests/test_claude_code_skills.py`
- [x] `uv run python -m pytest tests/test_cli_help.py tests/test_transcript.py`
- [x] reinstall the updated Claude Code plugin locally and verify the new skill surface is present in the runtime cache
- [x] invoke the Claude Code skill surface directly to confirm the router/session-recall contract is active


zc277584121:
Thanks for this substantial piece of work, @DarKWinGTM, and for the detailed review guide in #327. After thinking it through we're going to take a different direction and not merge this one.

## Why

memsearch is intentionally converging on a **single `memory-recall` skill with rich internal guidance** rather than a surface of several atomic skills + a router. Concretely, in the past few days we've been reinforcing that direction:

- #353 expanded the `memory-recall` description with explicit trigger examples ("what did I decide about X", "why did we do Y") so Claude knows when to invoke it
- #410 added a "When unsure what to search" section inside the skill pointing agents at `ls` / `grep` / `cat` of `.memsearch/memory/*.md` as a first-class exploration path
- #410 also replaced the SessionStart `tail -30` injection with a heading+bullet summary so agents see structure before ever having to form a search query

The underlying rationale is:

1. **One skill is easier to trigger correctly** than six. Claude already picks skills based on descriptions; having a router skill on top is a layer of indirection the runtime can largely handle on its own.
2. **The three-layer progressive disclosure (search → expand → transcript)** composes nicely inside a single skill invocation via `context: fork` — the sub-agent runs the whole chain and returns a curated summary, which keeps the main context clean.
3. **Adding 6 top-level skills widens the plugin's surface area** — more marketplace text to maintain, more trigger descriptions to keep aligned, more ways for users to accidentally invoke the wrong one.

That said, there are ideas in here we like (e.g. the tighter `session-recall` behavior that always tries `memsearch search "<session_id>"` before falling back to raw transcript reads — that's a good pattern worth folding into the `memory-recall` skill itself). If you want to open a smaller PR that adds session-ID-first recall to the existing `memory-recall` flow without introducing new top-level skills, we'd be happy to review.

Closing this and the companion review issue #327. Thanks again for the thoughtful proposal.

PR #308, opened 2026-04-09
fix: skip bm25 for tokenless query

haosenwang1018:
## Summary
- skip the BM25/sparse search leg when the query text has no lexical tokens to contribute
- avoid the Milvus Lite `Invalid sparse row: NaN or Inf value` crash reported for numeric-only queries like `memsearch search 111123`
- add a focused store test for the BM25 query-text normalization helper

## Testing
- git diff --check

### DIFF
--- a/src/memsearch/store.py
+++ b/src/memsearch/store.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import logging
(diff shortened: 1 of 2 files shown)


zc277584121:

PR #309, opened 2026-04-09
test: cover numeric-only search path

haosenwang1018:
## Summary
- add a store-level regression test for numeric-only search queries
- cover the user-facing `search ... --query-text 111123` path after the sparse/BM25 crash fix
- make sure the numeric-only query case stays non-crashing beyond the helper-only unit test

## Testing
- git diff --check


zc277584121:

PR #310, opened 2026-04-09
test: cover cli numeric-only search

haosenwang1018:
## Summary
- add a CLI regression test for numeric-only search queries
- cover the `memsearch search 111123` user path above the store-level crash fix
- ensure the CLI can invoke `MemSearch.search(...)` with numeric-only input and render results normally

## Testing
- uv run python -m pytest tests/test_search_cli.py -q
- git diff --check

### DIFF
--- a/src/memsearch/store.py
+++ b/src/memsearch/store.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import logging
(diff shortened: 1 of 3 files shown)


zc277584121:

PR #311, opened 2026-04-09
fix: keep alnum tech tokens in bm25

haosenwang1018:
## Summary
- keep BM25 enabled for technical query tokens that mix letters and digits, such as `gpt-4o`, `RFC-9110`, and `v2.1.0`
- avoid over-correcting the numeric-only query crash fix by treating these common technical identifiers as tokenless input
- extend the store test coverage for BM25 query normalization

## Testing
- git diff --check

### DIFF
--- a/src/memsearch/store.py
+++ b/src/memsearch/store.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import logging
+import re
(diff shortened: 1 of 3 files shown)


zc277584121:

PR #320, opened 2026-04-09
feat: add list memories command

haosenwang1018:
## Summary
- add `MemSearch.list_sources()` and a CLI `memsearch list` command for listing indexed memory sources
- provide a small answer to issue #277 without changing the search/index data model
- add CLI tests for both populated and empty indexes

## Testing
- uv run python -m pytest tests/test_search_cli.py -q
- git diff --check

### DIFF
--- a/plugins/claude-code/hooks/common.sh
+++ b/plugins/claude-code/hooks/common.sh
@@ -4,13 +4,25 @@
 
 set -euo pipefail
 
-# Read stdin JSON into $INPUT
-# Use timeout to prevent indefinite blocking in WSL 2 where stdin pipe may not close properly.
-# macOS lacks `timeout` — use perl alarm(2) as a portable fallback with a 2-second deadline.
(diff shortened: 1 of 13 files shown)


zc277584121:
Thanks for the PR. Closing this alongside #305 and #367 — after reviewing the underlying ask in #277, we decided not to add a `memsearch list` command.

Rationale and the alternative path we took are in the #305 closing comment. TL;DR: memsearch's "markdown is source of truth" design means `ls`/`grep`/`cat` of `.memsearch/memory/*.md` is strictly more powerful than a dedicated `list` command, and adding the command would leak the "chunk" abstraction.

What landed instead (PR #410): cold-start injection now surfaces heading structure instead of a raw tail, and the `memory-recall` SKILL.md files explicitly guide agents to explore the raw markdown when they don't know what to search for. That closes #277 without expanding the CLI surface.

Thanks again for the contribution.

PR #325, opened 2026-04-11
fix: derive memory-recall collections from git root

### LINKED ISSUE #324: [Bug] memory-recall skill uses wrong collection when launched from git subdirectory
## Bug Description

The `memory-recall` skill derives its collection name using `derive-collection.sh` **without arguments** (defaults to `pwd`), but `common.sh` (used by session-start hook) resolves the path via `git rev-parse --show-toplevel` first. When Claude Code is launched from a subdirectory of a git repo, these produce different collection names, causing the skill to search an empty collection.

## Reproduction Steps

1. Have a git repo at `/path/to/repo`
2. Launch Claude Code from a subdirectory: `cd /path/to/repo/some/subdir && claude`
3. Session-start hook indexes memory into collection `ms_repo_<hash_of_repo_root>`
4. Invoke `memory-recall` skill — SKILL.md runs `derive-collection.sh` with `pwd` — gets `ms_subdir_<hash_of_subdir>` (different collection, 0 chunks)
5. All searches return empty

## Root Cause

`skills/memory-recall/SKILL.md` line 12 uses no argument (defaults to `pwd`):

```
Collection: !`bash ${CLAUDE_PLUGIN_ROOT}/scripts/derive-collection.sh`
```

But `hooks/common.sh` uses git root:

```bash
_GIT_ROOT="$(git rev-parse --show-toplevel 2>/dev/null || echo "")"
if [ -n "$_GIT_ROOT" ]; then
  _PROJECT_DIR="$_GIT_ROOT"
else
  _PROJECT_DIR="${CLAUDE_PROJECT_DIR:-.}"
fi
COLLECTION_NAME=$(... derive-collection.sh "$_PROJECT_DIR" ...)
```

The skill doesn't pass git root to `derive-collection.sh`, while the hook does. This produces a different hash and a different (empty) collection.

## Suggested Fix

In `SKILL.md`, match `common.sh` logic — pass git root when available:

```
Collection: !`bash -c 'root=$(git rev-parse --show-toplevel 2>/dev/null || echo ""); if [ -n "$root" ]; then bash ${CLAUDE_PLUGIN_ROOT}/scripts/derive-collection.sh "$root"; else bash ${CLAUDE_PLUGIN_ROOT}/scripts/derive-collection.sh; fi'`
```

## Environment

- memsearch plugin version: 0.3.4
- Claude Code on macOS

haosenwang1018:
## Summary\n- make Claude Code memory-recall derive collections from the git repo root when available\n- apply the same repo-root derivation pattern to the Codex and OpenCode memory-recall skills\n- document the repo-root debugging path so manual checks match the hook-derived collection\n\n## Testing\n- from a nested directory, verify the skill command resolves the same collection name as plugins/claude-code/hooks/common.sh\n- git diff --check\n\n## Issue\n- fixes #324

### DIFF
--- a/docs/platforms/claude-code/memory-recall.md
+++ b/docs/platforms/claude-code/memory-recall.md
@@ -113,6 +113,12 @@ Manual invocation is useful when:
 
 **Edit memory files directly.** If a summary is inaccurate or contains sensitive information, open `.memsearch/memory/YYYY-MM-DD.md` in your editor and fix it. The watcher will re-index automatically. Memory files are plain markdown -- you're in full control.
 
+**Debug collection mismatches from subdirectories.** If Claude Code is launched from a git subdirectory, derive the collection from the repo root so manual debugging matches the hook and skill behavior:
+```bash
+root=$(git rev-parse --show-toplevel 2>/dev/null || pwd)
+bash /path/to/plugins/claude-code/scripts/derive-collection.sh "$root"
+```
+
 **Rebuild the index if search quality degrades.** If you change embedding providers or suspect index corruption:
 ```bash
 memsearch index .memsearch/memory/ --force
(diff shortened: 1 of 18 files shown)


zc277584121:
Thanks for catching this and for flagging that the Codex and OpenCode plugins share the same git-root collection bug — that was easy to miss.

The core fix has now landed in two smaller PRs:
- **#330** — Claude Code skill + docs (fixes #324)
- **#331** — Codex + OpenCode skills, plus `SessionEnd` marked `async: true`

With both merged, this PR is now superseded. Closing it here to keep the review queue clear. Appreciate the contribution!

Issue #221, opened 2026-03-23
[Bug&FR] VSCode Copilot Agent Hook Compability


KagaJiankui:
### Problem

当前提供的ccplugin不兼容VSCode Copilot提供的cc hook兼容层，会阻塞Copilot的回复导致Copilot完全无响应。禁用memsearch ccplugin后恢复正常。

Copilot Agent Hook日志如下：

```log
026-03-23 16:38:04.354 [info] [#2] [UserPromptSubmit] Executing 1 hook(s)
2026-03-23 16:38:04.354 [info] [#2] [UserPromptSubmit] Running: {"command":"bash c:\\Users\\Falke\\AppData\\Roaming\\Code\\agentPlugins\\github.com\\zilliztech\\memsearch\\ccplugin/hooks/user-prompt-submit.sh","cwd":{"$mid":1,"fsPath":"d:\\Electronics\\batt_data","_sep":1,"external":"file:///d%3A/Electronics/batt_data","path":"/d:/Electronics/batt_data","scheme":"file"},"env":{"CLAUDE_PLUGIN_ROOT":"c:\\Users\\Falke\\AppData\\Roaming\\Code\\agentPlugins\\github.com\\zilliztech\\memsearch\\ccplugin"},"timeout":15}
[...]

PR #326, opened 2026-04-12
fix: skip update checks during session start

haosenwang1018:
## Summary
- stop doing the PyPI update check in the default synchronous `SessionStart` hook path
- make update checks opt-in via `MEMSEARCH_HOOK_CHECK_UPDATES=1`
- reduce hook latency/cancellation risk for stricter runtimes such as VSCode Copilot's hook compatibility layer

## Related Issues
Part of #221

## Testing
- `bash -n ccplugin/hooks/session-start.sh ccplugin/hooks/common.sh ccplugin/hooks/user-prompt-submit.sh`
- `pytest -q tests/test_transcript.py`

### DIFF
--- a/plugins/claude-code/hooks/session-start.sh
+++ b/plugins/claude-code/hooks/session-start.sh
@@ -64,9 +64,13 @@ if [ -n "$REQUIRED_KEY" ] && [ -z "${!REQUIRED_KEY:-}" ]; then
   fi
 fi
 
-# Check PyPI for newer version (2s timeout, non-blocking on failure)
+# Optional PyPI update check.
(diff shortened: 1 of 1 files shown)


zc277584121:
Thanks for this contribution. #645 has now removed the repeated configuration reads reported in #632, but the synchronous PyPI update request remains a separate startup-cost concern, so this PR is not fully superseded. The SessionStart hook has changed substantially since this branch was opened and the branch now needs a rebase. Please rebase onto current main and add focused hook coverage for the default-disabled and opt-in update-check paths if you would like us to continue reviewing it.

Issue #354, opened 2026-04-14
`MEMSEARCH_DIR` in `common.sh` ignores environment variables.


marcoamtz:
### Problem

In `plugins/claude-code/hooks/common.sh:30-31`, `MEMSEARCH_DIR` is set via a straight assignment to `$_PROJECT_DIR/.memsearch`. This prevents users from centralizing their memory logs globally via an environment variable.

### Proposed solution

Change to `MEMSEARCH_DIR="${MEMSEARCH_DIR:-$_PROJECT_DIR/.memsearch}"` to allow for global overrides.

PR #355, opened 2026-04-15
fix(claude-code): allow MEMSEARCH_DIR override

haosenwang1018:
## Summary\n- allow Claude Code hook users to override  from the environment\n- keep the existing project-scoped  path as the default fallback\n\n## Testing\n- \n\nCloses #354


zc277584121:
Thanks for the contribution! Closing this as it's been superseded by #408, which was just merged. #408 covers both `plugins/claude-code/hooks/common.sh` and `plugins/codex/hooks/common.sh` with the same `${MEMSEARCH_DIR:-...}` pattern, so this PR's scope is already included on main.

Issue #352, opened 2026-04-14
In 0.3.0, the try/except block wrapping the `_on_change` file event callback in `core.py` was silently removed.


TadMSTR:
### Description

In 0.3.0, the try/except block wrapping the `_on_change` file event callback in `core.py` was silently removed. In 0.2.3, embedding errors during watch events were caught, logged, and skipped so the watcher kept running. In 0.3.0, those same errors propagate uncaught and crash the watcher process.

**0.2.3 (working):**
```python
def _on_change(event_type: str, file_path: Path) -> None:
    try:
        if event_type == "deleted":
            self._store.delete_by_source(str(file_path))
            summary = f"Removed chunks for {file_path}"
        else:
            n = loop.run_until_complete(self.index_file(file_path))
            summary = f"Indexed {n} chunks from {file_path}"
        logger.info(summary)
        if on_event is not None:
            on_event(event_type, summary, file_path)
    except Exception:
        logger.exception("Failed to index %s on change, skipping", file_path)
```
0.3.0: The try/except is gone — exceptions propagate to the caller and crash the watcher.
Any transient or file-specific embedding error (network blip, model 500, rate limit) now terminates the watcher entirely instead of logging the error and continuing.

### Steps to reproduce

1. Configure memsearch with an Ollama embedding provider
2. Have any file in the watched directory that causes the embedding provider to raise an exception (e.g., an Ollama 500 error for a specific file)
3. Run memsearch watch
4. The watcher crashes on first encounter of that file instead of logging and skipping it

### Error output / logs

```shell
`ollama._types.ResponseError: failed to encode response: json: unsupported value: NaN (status code: 500)`
```

### memsearch version

0.3.0

### Operating system

Linux

### Python version

3.13.2

Issue #102, opened 2026-03-04
Enhance robustness of memory search with jieba-next CJK tokenizer and native Markdown parsing


KagaJiankui:
## Problem Statement

There are two **severe** design flaws that   carries  a high risk of compromising   information retrieval effectiveness and degrading overall reliability, especially in CJK (specifically Chinese) contexts:

- The current lack of robust **tokenization** and **semantic parsing** forms a critical bottleneck for practical usage in multilingual (notably CJK) environments.
- LLMs may produce malformed or inconsistent Markdown output: their markdown *dialect* is shaped by black-box pre-training/RLHF, making formats unstable even with explicit prompts. And infrastructure issues (network fluctuations, cluster load balancing) often cause errors in complex structures like tables.
- Plain regex chunking and standard tokenizers (as in current store.py/chunker.py) lead to poor recall, fuzzy chunking, and ineffective BM25 search in Chinese.

## Proposal: Tokenizer and Markdown Parser Recommendations

- **Chinese tokenizer:** Prioritize [jieba-next](https://github.com/mxcoras/jieba-next) (most API-rich, good Windows support) for CJK tokenization. Alternatively, [jieba-rs](https://github.com/messense/jieba-rs) offers higher speed but narrower API coverage for basic use cases.
- **Markdown parser:** Favor [Mistune](https://github.com/lepture/mistune), as it is actively maintained and provides a structured AST in Python. Alternatively, [md4c](https://github.com/mity/md4c) is very fast, but the SAX parser design can be fragile in engineering for precise chunking.

## Potential Consequences

Failure to address these design flaws will:
- Severely limit semantic search quality, especially for Chinese/CJK agents or knowledge bases
- Hinder adoption as a memory solution in multilingual/global settings
- Brittle heading/paragraph detection increases risk of false negatives/positives during recall

**Request:** Please consider prioritizing a robust chunker and BM25 tokenizer/pluggable analyzer as a core architectural improvement.

chenhz01:
[...]

PR #388, opened 2026-04-15
fix(chunker): avoid splitting engineering text on ascii dots

haosenwang1018:
## Summary
- only treat ASCII sentence punctuation as a boundary when it looks like a real sentence ending
- avoid splitting inside emails, URLs, file extensions, and similar engineering text
- preserve mixed English-to-CJK sentence boundaries when punctuation is followed directly by CJK text

## Testing
- uv run python -m pytest tests/test_chunker.py

Part of #102

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -1,6 +1,6 @@
 # Architecture
 
-This page explains the technical architecture and key implementation decisions behind memsearch. For design principles, competitor comparison, and the "why" behind these decisions, see [Design Philosophy](design-philosophy.md).
+This page explains the technical architecture and key implementation decisions behind memsearch. New here? Start with [Getting Started](getting-started.md) for the end-to-end user flow before diving into internals. For design principles, competitor comparison, and the "why" behind these decisions, see [Design Philosophy](design-philosophy.md).
 
 ---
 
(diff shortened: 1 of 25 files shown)


zc277584121:
Closing this as all four parts of the bundle have now landed on main through focused PRs:

- **Chunker engineering-text regex** (the main value of this PR) — now on main as #412. We cherry-picked the `_SENTENCE_END_RE` change with a tighter lookahead form and added 6 targeted tests (CJK full-stop, mixed CJK/English, ASCII `. ` still splits prose, emails/URLs/file-paths never split mid-token, ellipsis, fullwidth semicolon).
- **`_on_change` try/except for watch callback** → covered by #336 + #409 (broadened to `except Exception` to actually fix #352's Ollama NaN repro).
- **`memsearch list` CLI command** → after revisiting #277 we decided not to add this CLI surface at all; instead we improved the SessionStart cold-start injection (heading+bullet summary) and the `memory-recall` skill to point agents at `ls`/`grep`/`cat` of `.memsearch/memory/*.md` (#410). See the #277 closing comment for the full rationale.
- **`store.query` missing-collection tolerance** → already tolerated by all current callers; didn't need a dedicated code path.

Thanks for the chunker regex in particular — that was a real bug and the fix shipped. Closing this bundle now that the individual pieces are upstream.

PR #506, opened 2026-04-25
feat(opencode): add turn-level transcript recall

### LINKED ISSUE #505: [OpenCode Plugin] Add turn-level cursor to `memory_transcript` for precise L3 retrieval
## Problem
  The OpenCode L3 `memory_transcript` path can only retrieve a tail window of a session (`--limit`) or the last turn (`--last-turn`). It cannot jump to an arbitrary turn in a long session,
  so transcript drill-down becomes coarse and inefficient as sessions grow.

  ## Current behavior
  - `plugins/opencode/scripts/parse-transcript.py` loads messages by `session_id` ordered by `time_created`.
  - The script exposes `--limit` and `--last-turn`, but no stable turn/message cursor.
  - The OpenCode capture flow stores daily markdown summaries, but does not persist a reusable turn anchor for later drill-down.

  ## Why this matters
  For long sessions, L3 recall often needs to land on the exact turn that introduced a decision, error, or code snippet. Without a turn-level cursor, the tool has to scan too much
  surrounding text and can return more irrelevant context than needed.

  ## Reference
  The Claude Code plugin already writes turn-level anchors and supports drilling into a specific turn with surrounding context via `memsearch transcript ... --turn ... --context ...`.

## Proposed change
  Add a stable cursor to the OpenCode capture/transcript flow, for example:
  - store a turn/message identifier in the anchor
  - allow `memory_transcript` to accept `--turn` or `--message-id`
  - add a `--context` option so the tool can return the target turn plus surrounding messages

  The exact cursor field can match OpenCode's schema, as long as it reliably identifies a turn and its nearby context.

  ## Expected result
  `memory_transcript` can jump directly to the relevant turn instead of only reading a session tail.

### Alternatives considered

_No response_

InSciTech:
Closes #505

## Summary
- add turn anchors and `turn_id`/`context` transcript drill-down for the OpenCode plugin
- keep transcript recall sourced from the OpenCode SQLite DB while treating `.memsearch/opencode-turns.db` as derived capture state for checkpoints and stable replay ordering
- clarify markdown-first replay semantics and add regression coverage for partial sidecar failures

## Test Plan
- `uv run python -m pytest tests/test_opencode_turns.py tests/test_opencode_parse_transcript.py -v`
- `uv run ruff check tests/test_opencode_turns.py tests/test_opencode_parse_transcript.py plugins/opencode/scripts/parse-transcript.py plugins/opencode/scripts/opencode_turns.py plugins/opencode/scripts/capture-daemon.py`

### DIFF
--- a/docs/platforms/opencode/how-it-works.md
+++ b/docs/platforms/opencode/how-it-works.md
@@ -53,7 +53,7 @@ OpenCode stores all conversations in a SQLite database (`~/.local/share/opencode
 
 - **No hook limitations** -- capture works regardless of which hooks OpenCode exposes
 - **Reliable detection** -- new turns are detected by tracking a per-session completed-turn cursor, not by fragile event timing
-- **Crash resilience** -- state is persisted to `.memsearch/opencode-turns.db`, so daemon restarts don't re-capture old turns
+- **Crash resilience** -- derived state is persisted to `.memsearch/opencode-turns.db`, so daemon restarts can replay safely without duplicating captured markdown
 
 ### Daemon Flow
 
@@ -86,8 +86,13 @@ Step by step:
 3. **Extract text** -- reads message `parts` (text content, tool calls with names/paths) into a readable format
 4. **Summarize** -- calls `opencode run` with the turn text and a third-person summarization prompt
 5. **Write to memory** -- appends the summary to `.memsearch/memory/YYYY-MM-DD.md` with `<!-- session:ID turn:ID db:PATH -->` anchors
-6. **Persist state** -- writes the completed-turn cursor to `.memsearch/opencode-turns.db` so restarts don't re-capture
+6. **Persist state** -- writes the completed-turn cursor and derived turn ordering to `.memsearch/opencode-turns.db`
 7. **Re-index** -- triggers `memsearch index` in the background
+
+OpenCode SQLite remains the source of truth for original transcript reads. The
+sidecar database is derived capture state only. If markdown append succeeds and
+the sidecar write fails later, the next daemon replay uses the existing
+session+turn anchor to avoid duplicate memory entries and repair the sidecar.
 
 ### LLM Summarization with Isolation
 
@@ -180,7 +185,7 @@ your-project/.memsearch/memory/
 - Added structured error logging with request ID correlation
 ```
 
-The `<!-- session:... turn:... db:~/.local/share/opencode/opencode.db -->` anchors are used by the `memory_transcript` tool to query the original conversation from OpenCode's SQLite database.
+The `<!-- session:... turn:... db:~/.local/share/opencode/opencode.db -->` anchors are used by the `memory_transcript` tool to query the original conversation from OpenCode's SQLite database. The sidecar database is not required for transcript reads.
 
 ---
 
--- a/docs/platforms/opencode/how-it-works.md
+++ b/docs/platforms/opencode/how-it-works.md
@@ -52,8 +52,8 @@ Unlike Claude Code and Codex (which use hook-based capture), OpenCode uses a **b
 OpenCode stores all conversations in a SQLite database (`~/.local/share/opencode/opencode.db`). The daemon polls this database directly, which means:
 
 - **No hook limitations** -- capture works regardless of which hooks OpenCode exposes
-- **Reliable detection** -- new turns are detected by tracking `last_msg_time`, not by fragile event timing
-- **Crash resilience** -- state is persisted to `.memsearch/.last_msg_time`, so daemon restarts don't re-capture old turns
+- **Reliable detection** -- new turns are detected by tracking a per-session completed-turn cursor, not by fragile event timing
+- **Crash resilience** -- state is persisted to `.memsearch/opencode-turns.db`, so daemon restarts don't re-capture old turns
 
 ### Daemon Flow
 
@@ -67,13 +67,13 @@ sequenceDiagram
 
     loop Every 10 seconds
         Daemon->>DB: Query sessions for project_dir
-        DB->>Daemon: Messages newer than last_msg_time
+        DB->>Daemon: Messages newer than the last completed turn cursor
         alt New turns found
-            Daemon->>Daemon: Group into user+assistant pairs
+            Daemon->>Daemon: Group into turns via user message + assistant descendants
             Daemon->>LLM: Summarize turn (isolated config)
             LLM->>Daemon: 2-6 bullet points
             Daemon->>File: Append with session anchor
-            Daemon->>Daemon: Update last_msg_time (persist to disk)
+            Daemon->>Daemon: Update turn cursor (persist to sidecar DB)
             Daemon->>Index: memsearch index (background)
         end
     end
@@ -81,12 +81,12 @@ sequenceDiagram
 
 Step by step:
 
-1. **Poll SQLite** -- queries the `session` and `message` tables for the current project directory, looking for messages newer than `last_msg_time`
-2. **Group into turns** -- pairs consecutive `user` + `assistant` messages into turns
+1. **Poll SQLite** -- queries the `session` and `message` tables for the current project directory, looking for messages newer than the last completed turn cursor
+2. **Group into turns** -- groups each `user` message with its assistant/tool descendants until the next `user` message
 3. **Extract text** -- reads message `parts` (text content, tool calls with names/paths) into a readable format
 4. **Summarize** -- calls `opencode run` with the turn text and a third-person summarization prompt
-5. **Write to memory** -- appends the summary to `.memsearch/memory/YYYY-MM-DD.md` with `<!-- session:ID db:PATH -->` anchors
-6. **Persist state** -- writes `last_msg_time` to `.memsearch/.last_msg_time` so restarts don't re-capture
+5. **Write to memory** -- appends the summary to `.memsearch/memory/YYYY-MM-DD.md` with `<!-- session:ID turn:ID db:PATH -->` anchors
+6. **Persist state** -- writes the completed-turn cursor to `.memsearch/opencode-turns.db` so restarts don't re-capture
 7. **Re-index** -- triggers `memsearch index` in the background
 
 ### LLM Summarization with Isolation
@@ -97,8 +97,8 @@ The daemon summarizes turns via `opencode run` -- but it must avoid triggering t
 result = subprocess.run(
     ["opencode", "run", "-m", small_model, prompt],
     env={
-        "XDG_CONFIG_HOME": "/tmp/opencode-memsearch-summarize",
-        "XDG_DATA_HOME": "/tmp/opencode-memsearch-summarize/data",
+        "XDG_CONFIG_HOME": "~/.codex/tmp/opencode-memsearch-summarize",
+        "XDG_DATA_HOME": "~/.codex/tmp/opencode-memsearch-summarize/data",
         "MEMSEARCH_NO_WATCH": "1",
     },
 )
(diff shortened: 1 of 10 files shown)


zc277584121:
Thanks for putting this together. The issue makes sense to me: OpenCode L3 recall currently only has session-level drill-down, so long sessions need a stable turn cursor.

I have one design question before merging this: is `.memsearch/opencode-turns.db` intended to be required for turn-level transcript recall, or primarily a derived capture checkpoint/cache?

From reading the PR, the minimal L3 path seems to be:

- write `turn:<user_message_id>` into the markdown anchor during capture
- have `memory_transcript(session_id, turn_id, context)` rebuild turns from the OpenCode SQLite DB on demand
- use the target turn id to return surrounding context

That would solve the transcript drill-down problem without introducing another sidecar DB. The sidecar DB does seem useful for replacing `.last_msg_time` with a more robust per-session/per-turn capture checkpoint, but that is a bigger state-management change.

If we keep the sidecar DB in this PR, I think we should make a few things explicit:

- document that `.memsearch/opencode-turns.db` is derived state only, not memory source of truth; it should be safe to delete/rebuild
- define what happens if sidecar state and markdown disagree, e.g. markdown write succeeds but state update fails, or state update succeeds but markdown write fails
- add/confirm tests for those recovery paths
- fix the current ruff issue in `tests/test_opencode_turns.py` (`SIM105`)
- check whether the mode changes from `100755` to `100644` for `capture-daemon.py` and `parse-transcript.py` are intentional; current TS calls use `python3`, so it may be harmless, but it looks like unnecessary metadata churn unless direct execution is no longer supported

Overall I like the direction, but I would prefer we treat the sidecar DB as a deliberate capture-checkpoint design change, not just an implementation detail of `memory_transcript`.

InSciTech:
You're reading the branch correctly: `.memsearch/opencode-turns.db` is not required for `memory_transcript`.

On the design question: transcript reads still rebuild turns from the OpenCode SQLite database on demand. The sidecar is derived capture state only: replay-safe checkpoints and stable turn ordering for the daemon. I also updated the PR description so it no longer frames the sidecar as part of the transcript read path.

On the recovery-path point: the current write order is markdown append -> `save_turn()` -> `save_turn_state()`. So the meaningful replay cases here are:
- markdown write succeeded, `save_turn()` failed
- markdown write succeeded, `save_turn_state()` failed
- markdown write failed before any sidecar state advanced

This update adds or confirms coverage for those paths:
- `test_capture_session_turns_repairs_sidecar_after_partial_turn_save_failure`
- `test_capture_session_turns_is_idempotent_after_partial_state_save_failure`
- `test_capture_session_turns_does_not_advance_sidecar_when_markdown_write_fails`

On the source-of-truth point: I clarified in code and docs that:
- OpenCode SQLite is the source of truth for transcript recall
- `.memsearch/opencode-turns.db` is rebuildable derived state only
- replay dedupes through existing `session + turn_id` markdown anchors

I also added `test_parse_transcript_reads_from_opencode_sqlite_without_sidecar` to make the non-dependence of transcript reads on the sidecar explicit.

On the other review items:
- fixed the `SIM105` lint issue in `tests/test_opencode_turns.py`
- restored the unintentional `100755` script modes for `capture-daemon.py` and `parse-transcript.py`

If you'd still prefer the sidecar/checkpoint work to be split after this clarification, I can do that next.

zc277584121:
Thanks for the update. The clarification that transcript reads still rebuild from the OpenCode SQLite DB, and that `opencode-turns.db` is derived capture state only, addresses my main design concern. I also pulled the latest branch locally and the focused pytest/ruff checks pass for me.

I did one more pass and found two edge cases I think are worth handling before merge:

1. Upgrade / migration from the existing `.last_msg_time` state

This PR switches capture progress to the new sidecar state. For an existing user who already has `.memsearch/.last_msg_time` and old markdown anchors like:

```md
<!-- session:... db:... -->
```

but no `.memsearch/opencode-turns.db` yet, the new daemon appears to start with an empty `turn_state`, so `after_time` is `None`. Since `capture_exists()` only dedupes new `session + turn` anchors, old captured entries without `turn:` would not be recognized as already captured.

That could cause an upgrade to replay/re-summarize recent OpenCode session history and append duplicate memory entries. Could you add either a small migration path from `.last_msg_time`, or an explicit first-run behavior/test that avoids duplicate capture for existing installs?

2. Textless assistant/tool-call messages in the parent chain

`build_turns()` currently skips any message whose rendered text is empty before adding its id to `current_message_ids`. That can break descendant grouping if OpenCode stores an intermediate assistant/tool-call message with no text but a later assistant message points to it via `parentID`.

For example:

```text
u1 user text
  a1 assistant, finish=tool-calls, no text
    a2 assistant, parentID=a1, text="final answer"
```

In this shape, `a1` is skipped, so `a2`'s parent is not in `current_message_ids` and the turn remains incomplete. I think the turn builder should preserve structural assistant message ids even when their rendered text is empty, or otherwise special-case this parent-chain scenario. A regression test for this would be useful.

Other than those two points, the updated direction looks good to me.

InSciTech:
I've pushed follow-ups for both remaining edge cases.

1. Existing-install upgrade from `.last_msg_time`

- The sidecar `turn_state` remains the primary checkpoint source.
- If the sidecar has not established any cursor yet, `capture_session_turns()` now falls back to the legacy `.memsearch/.last_msg_time` value as the `after_time` cutoff.
- This prevents an existing install with old markdown anchors and no `opencode-turns.db` from replaying pre-sidecar history on the first upgraded daemon run.
- Once a new turn is captured after the upgrade, the new `turn_state` row is written and subsequent runs continue from the sidecar normally.
- I also added regression coverage for this exact upgrade shape: old `.last_msg_time` + old `session/db` anchor + empty sidecar -> only turns after the legacy cutoff are captured, and the new sidecar cursor is seeded from the first newly captured turn.

2. Textless assistant / tool-call parent chain

- `build_turns()` no longer drops structural assistant nodes just because their rendered text is empty.
- When an assistant message belongs to the current turn but renders empty, its message id is still preserved in the parent chain so descendant assistant messages can attach to it.
- The empty node is kept as a structural edge only; it is not appended to `turn.messages`, so rendered output and message-count semantics stay stable.
- I added a regression test for the `u1 -> a1 (tool-calls, no text) -> a2 (final answer)` shape to ensure the final assistant reply stays in the same turn.

One nuance on the upgrade path: this deliberately does not rewrite older markdown anchors. Older memories without `turn:` continue to drill down through the existing session-level `memory_transcript(... --limit ...)` fallback, while newly captured memories enter the turn-level drill-down path with `turn:` anchors. The goal of this change is to make the upgrade safe and non-duplicating, not to retroactively remap historical markdown.

Verification run locally before pushing:

- `uv run python -m pytest tests/test_opencode_turns.py -v`
- `uv run ruff check tests/test_opencode_turns.py plugins/opencode/scripts/opencode_turns.py plugins/opencode/scripts/capture-daemon.py`

InSciTech:
I've force-pushed a rebuilt branch history so the PR now reads as five English commits, and the branch now includes the two review fixes plus the tail-turn capture correction that I want to carry with this turn-based design.

The branch now breaks down into:

1. `feat(opencode): add turn-level transcript recall`
2. `fix(opencode): clarify sidecar transcript boundaries`
3. `fix(opencode): preserve assistant parent chains through textless nodes`
4. `fix(opencode): preserve legacy .last_msg_time on sidecar upgrade`
5. `fix(opencode): delay tail-turn capture until content stays stable`

The first four commits address the original turn-level recall work plus the two review edge cases:

- textless assistant / tool-call bridge messages no longer break descendant attachment inside a turn
- existing installs with only `.last_msg_time` and old `session/db` anchors no longer replay pre-sidecar history on first upgrade

I also chose to carry one more correctness fix that is specific to this turn-based capture design:

## Tail-turn capture timing

The original turn-complete heuristic still treated the current tail turn too optimistically. In long-running search/tool/synthesis flows, the assistant can temporarily stop looking like `tool-calls`, then continue appending output. That meant the daemon could summarize and persist a half-finished tail turn into memory markdown before the content had really stabilized.

The updated behavior is now:

- if a later user message exists, the previous turn closes immediately
- otherwise, the current tail turn must remain structurally/textually unchanged for 300 seconds before capture
- any content change resets the timer

The fingerprint used for that quiet-period check is based on the actual turn shape and rendered content rather than only the latest message id, so continued search output, additional assistant text, or other structural changes keep the tail turn open.

I added regression coverage for:

- waiting for the 300s quiet period before capturing the tail turn
- resetting the quiet period when the tail turn content changes
- immediately closing the previous turn once the next user message arrives

## Scope / source-of-truth note

This does **not** change the transcript source of truth: `memory_transcript` still rebuilds from the OpenCode SQLite database on demand.

It also does **not** rewrite historical markdown anchors during upgrade:

- older memories without `turn:` continue to use the existing session-level transcript fallback
- newly captured memories continue entering the turn-level drill-down path with `turn:` anchors

## Verification run before the force-push

- `uv run python -m pytest tests/test_opencode_turns.py tests/test_opencode_parse_transcript.py -v`
- `uv run ruff check tests/test_opencode_turns.py tests/test_opencode_parse_transcript.py plugins/opencode/scripts/opencode_turns.py plugins/opencode/scripts/capture-daemon.py plugins/opencode/scripts/parse-transcript.py`

If you'd prefer, I can also split the tail-turn capture-timing change back out into a follow-up PR, but I believe it belongs with this turn-based capture path because it fixes when the new turn model is allowed to persist the current tail turn at all.

zc277584121:
Thanks for the detailed follow-ups. The latest direction looks much closer to mergeable to me.

The important design boundary is now clear: OpenCode SQLite remains the transcript source of truth, while `.memsearch/opencode-turns.db` is derived capture/checkpoint state. The added coverage around transcript reads without the sidecar, legacy `.last_msg_time` upgrade behavior, textless assistant parent chains, and tail-turn stability addresses the main concerns I had in the previous review.

I think there are only two remaining things to clean up before merge:

1. Please rebase this branch onto the latest `main` and resolve the current `capture-daemon.py` conflict without dropping the newer main-branch fixes, especially:
   - quoted `memsearch_cmd` handling via `shlex.split()`
   - `plugins.opencode.summarize.model` override support
   - the current isolated config handling for summarization

2. Please move the existing-anchor check before summarization in the replay path. If `.memsearch/opencode-turns.db` is deleted and rebuilt, turns that already have markdown anchors should repair sidecar state without calling the summarizer again. A small regression test for that rebuild case would be useful.

With those handled, I think this PR is aligned with the plugin design and should be in good shape to merge.

InSciTech:
Thanks for the review. I've pushed updates addressing both remaining items.

**1. Rebase onto latest `main` with conflict resolution**

The branch is now rebased onto `origin/main` (which includes the new `shlex.split()`, `get_plugin_summarize_model()`, and symlink cleanup from `7a0307c`). The `capture-daemon.py` conflict was resolved by keeping the turn-level architecture while incorporating all three main-branch fixes:

- `split_memsearch_cmd()` via `shlex.split()` — replaces `memsearch_cmd.split()` in both `_load_summarize_prompt()` and `get_plugin_summarize_model()`
- `get_plugin_summarize_model()` — reads `plugins.opencode.summarize.model` config override, used in `summarize_with_llm()` as `summarize_model = get_plugin_summarize_model(memsearch_cmd) or small_model`
- Symlink/broken-dst cleanup in `ensure_isolated_config()` via `contextlib.suppress`

No main-branch fixes were dropped.

**2. Move existing-anchor check before summarization**

The replay path now checks `capture_exists()` before calling `summarize_with_llm()`. When `opencode-turns.db` is deleted and rebuilt, turns that already have markdown anchors repair sidecar state (`save_turn()` + `save_turn_state()`) without calling the summarizer again.

The updated logic in `capture_session_turns()`:

```python
# Before (wasteful on rebuild):
summary = summarize_with_llm(turn_text, small_model, memsearch_cmd)
if not capture_exists(memory_dir, session_id, turn.turn_id):
    write_capture(...)

# After (anchor check first):
if not capture_exists(memory_dir, session_id, turn.turn_id):
    summary = summarize_with_llm(turn_text, small_model, memsearch_cmd)
    write_capture(...)
save_turn(turn_db, turn)
```

Added regression test: `test_capture_session_turns_skips_summarizer_when_anchor_already_exists` — verifies that summarizer call count stays at 1 when replaying a session whose anchors are already written.

**Verification:**

- `uv run python -m pytest tests/test_opencode_turns.py tests/test_opencode_parse_transcript.py -v` — 13 passed
- `uv run ruff check tests/test_opencode_turns.py tests/test_opencode_parse_transcript.py plugins/opencode/scripts/opencode_turns.py plugins/opencode/scripts/capture-daemon.py plugins/opencode/scripts/parse-transcript.py` — All checks passed!

Issue #337, opened 2026-04-14
Feature request: recommended personal/global memory alongside project-scoped memory


congmucc:
### Problem

`memsearch`'s project-scoped memory model is a strong default. The per-project `.memsearch/memory/` directory and per-project collection derivation make isolation clear and predictable, which works very well for project history.

However, in real usage there is another class of memory that does not really belong to a single project:

- stable user preferences
- coding style preferences
- preferred tools/workflows
- recurring collaboration preferences
- durable "how I like to work" context

For agent workflows, a recommended two-layer model often feels natural:

1. Project memory
   - architecture decisions
   - debugging history
   - project-specific conventions
   - task history

2. Personal/global memory
   - user preferences that should carry across projects
   - general coding preferences
   - preferred output style / collaboration style
   - durable habits that are not tied to one repo

Project-only memory is still a very good default. The gap is that personal/global memory does not feel like a first-class or officially recommended pattern today, so users end up duplicating the same preferences into many projects or building custom workarounds outside the normal flow.

### Proposed solution

I would love to see an officially recommended personal/global memory layer alongside the existing project-scoped model.

I am not asking for this to be mandatory, and I am not asking for a large multi-bank platform.

A small optional model would already help a lot:

- keep `project` memory exactly as the default
- optionally support `personal` / `global` memory as a second scope

### Possible shape

Storage:

- keep current project storage exactly as-is
- add a separate personal/global memory location, for example:
  - `~/.memsearch/personal/memory/`
  - or another configurable global path

Retrieval:

- search current project memory first
- optionally merge in personal/global memories
- show which result came from which scope
- keep project memory higher priority

Capture:

- project memory stays the default capture target
- personal/global memory can be opt-in
- even a minimal first version would help, for example:
  - explicit/manual capture
  - a simple tag/frontmatter-based routing mechanism
  - or a lightweight configurable rule for durable user preferences

### Why this would help

This would preserve the current strengths of `memsearch`:

- markdown remains the source of truth
- project isolation remains strong by default
- advanced users get a clean path for durable cross-project preferences

From my perspective, the key ask is not "force everyone into multiple scopes". The key ask is to make this a documented and supported recommended pattern for users who want both:

- strong per-project memory
- stable cross-project personal memory

### Alternatives considered

Current workarounds seem to be:

- duplicating personal preferences into each project
- relying only on agent global instructions outside memsearch
- building a custom second collection / second memory directory manually

[...]

PR #514, opened 2026-04-28
fix(plugins): derive collection from MEMSEARCH_DIR when explicitly set

astefanutti:
## Summary

When `MEMSEARCH_DIR` is explicitly set via environment variable (e.g. for cross-project global memory), the Milvus collection name is now derived from `MEMSEARCH_DIR` instead of the project directory. This ensures that both the markdown files **and** the vector index share the same scope.

**Problem:** PR #408 added support for `MEMSEARCH_DIR` override, which correctly routes markdown files to a shared directory. However, the collection name is still derived from `$_PROJECT_DIR` (git root), so memories written in Project A are not searchable from Project B — the vector index remains per-project even though the files are global.

**Fix:** Check whether `MEMSEARCH_DIR` was explicitly set before the default assignment. When explicit, derive the collection from `MEMSEARCH_DIR`. Default behavior (no env var) is unchanged.

Applied to both Claude Code and Codex plugins.

Relates to #337

## Test plan

- [ ] Set `export MEMSEARCH_DIR="$HOME/.memsearch/global"` in shell profile
- [ ] Start a Claude Code session in Project A, generate a memory
- [ ] Start a Claude Code session in Project B, search for the memory from Project A
- [ ] Verify the same collection name is used in both projects
- [ ] Without `MEMSEARCH_DIR` set, verify per-project isolation still works (different collection names)

### DIFF
--- a/plugins/claude-code/hooks/common.sh
+++ b/plugins/claude-code/hooks/common.sh
@@ -27,6 +27,9 @@ if [ -n "$_GIT_ROOT" ]; then
 else
   _PROJECT_DIR="${CLAUDE_PROJECT_DIR:-.}"
 fi
+# When MEMSEARCH_DIR is explicitly set, use global scope (shared dir + collection).
+# Otherwise, default to per-project isolation.
+_MEMSEARCH_DIR_EXPLICIT="${MEMSEARCH_DIR:+true}"
 MEMSEARCH_DIR="${MEMSEARCH_DIR:-$_PROJECT_DIR/.memsearch}"
 MEMORY_DIR="$MEMSEARCH_DIR/memory"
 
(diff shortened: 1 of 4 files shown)


zc277584121:
Thanks, this direction makes sense to me. The hooks-side change covers capture/indexing, but I think the recall skills need the same collection derivation rule as well.

Right now, with `MEMSEARCH_DIR` set, the hooks index into a collection derived from `MEMSEARCH_DIR`, while the Claude Code and Codex memory-recall skills still derive the collection from the git/project root. That means capture and recall can point at different collections.

Could you also update:

- `plugins/claude-code/skills/memory-recall/SKILL.md`
- `plugins/codex/skills/memory-recall/SKILL.md`

to prefer `MEMSEARCH_DIR` when it is explicitly set, and fall back to the project root otherwise?

The fallback raw markdown paths in those skills should probably follow the same memory dir too.

astefanutti:
@zc277584121 thanks a lot for the quick review.

I've added be9d2bd40b3308acd9b58b3b0d2b82e87ee072c5 that should implement your feedback.

PR #543, opened 2026-05-17
fix(store): flush remote upserts for immediate search

dommonkhouse:
## Summary
- Flush remote Milvus collections after upsert so newly captured chunks are immediately searchable
- Keep Milvus Lite behaviour unchanged
- Add regression coverage for remote and local upsert behaviour

## Verification
- uv run pytest tests/test_store.py
- uv run ruff check src/memsearch/store.py tests/test_store.py
- git diff --check

Note: full suite currently has one unrelated failure in tests/test_embeddings_openai.py::test_embed_deterministic due exact float comparison drift.

### DIFF
--- a/src/memsearch/store.py
+++ b/src/memsearch/store.py
@@ -134,6 +134,8 @@ class MilvusStore:
             collection_name=self._collection,
             data=chunks,
         )
+        if not self._is_lite:
+            self._client.flush(self._collection)
         return result.get("upsert_count", len(chunks)) if isinstance(result, dict) else len(chunks)
 
     def search(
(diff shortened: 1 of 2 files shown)


dommonkhouse:
Fresh reproduction from a remote Milvus setup using upstream vanilla `memsearch[onnx]==0.4.3`:\n\n- Temporary collection: `ms_vanilla_probe_<timestamp>`\n- `memsearch index` reported `Indexed 1 chunks.`\n- Immediate `memsearch stats` for that collection reported `Total indexed chunks: 0`\n- Searches through the agreed 5-second window returned `[]`\n- The same test shape using my fork with this PR's flush patch reported `Total indexed chunks: 1` immediately and search found the probe at poll 0\n\nSo Milvus auto-flush is not enough for the write-then-search workflow on remote Milvus; an explicit flush after upsert is still needed for immediate visibility.

dommonkhouse:
Ready for maintainer review and merge. I verified this PR is non-draft, mergeable, and has no configured checks reporting failures.

PR #561, opened 2026-05-29
fix(codex): preserve explicit memory directory in installer

dommonkhouse:
## Summary
- preserve an explicitly supplied MEMSEARCH_DIR in installed Codex hook commands
- keep the default installer behaviour unchanged when MEMSEARCH_DIR is not set
- quote the preserved path safely for hook command execution

## Verification
- bash -n plugins/codex/scripts/install.sh
- installer smoke test with MEMSEARCH_DIR unset writes normal hook commands
- installer smoke test with MEMSEARCH_DIR set writes MEMSEARCH_DIR-prefixed hook commands
- git diff --check
- scanned diff additions for user-specific local paths

### DIFF
--- a/plugins/codex/scripts/install.sh
+++ b/plugins/codex/scripts/install.sh
@@ -67,16 +67,19 @@ PY
 install_or_update_hooks_file() {
   local hooks_file="$1"
   local install_dir="$2"
+  local memsearch_dir="${3:-}"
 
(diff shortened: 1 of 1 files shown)


dommonkhouse:
Ready for maintainer review and merge. I verified this PR is non-draft, mergeable, and has no configured checks reporting failures.

PR #564, opened 2026-06-01
fix(codex): preserve UTF-8 in stop summaries

valbarko:
## Summary
- replace byte-based truncation in the Codex stop hook with Unicode-safe character truncation
- sanitize fallback summary output before appending to memory markdown
- add a regression test that runs the stop worker with long Cyrillic text and reads the resulting memory file as UTF-8

## Verification
- bash -n plugins/codex/hooks/stop.sh
- uv run pytest tests/test_codex_stop_hook_utf8.py tests/test_codex_parse_rollout.py
- uv run ruff check tests/test_codex_stop_hook_utf8.py
- git diff --check

### DIFF
--- a/plugins/codex/hooks/stop.sh
+++ b/plugins/codex/hooks/stop.sh
@@ -48,6 +48,20 @@ _load_worker_field() {
   _json_val "$work_input" "$key" ""
 }
 
+_truncate_chars() {
+  local limit="$1"
+  python3 -c "
+import sys
+limit = int(sys.argv[1])
+text = sys.stdin.read()
+sys.stdout.write(text[:limit])
+" "$limit" 2>/dev/null || true
+}
+
+_valid_utf8() {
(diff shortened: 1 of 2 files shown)


plugins/codex/hooks/stop.sh
```diff
@@ -48,6 +48,20 @@ _load_worker_field() {
   _json_val "$work_input" "$key" ""
 }
 
+_truncate_chars() {
```
googs1025, line 51: I hit this issue locally with the Codex hook as well. The byte-oriented truncation path can leave invalid UTF-8 in the markdown memory file, and then later
  `memsearch index .memsearch/memory` reports a UnicodeDecodeError when reading that file.

googs1025:
This PR looks like the right source-side fix for Codex:
  - character-based truncation avoids splitting multi-byte UTF-8 sequences
  - the regression test exercises the worker fallback path with non-ASCII text

PR #607, opened 2026-07-05
feat: support ONNX models that require token_type_ids

j-sperling:
## Summary
- BERT-family ONNX exports declare a `token_type_ids` input, and `session.run()` requires every declared input in the feed — so models like `Xenova/all-MiniLM-L6-v2` and `Xenova/bge-small-en-v1.5` currently fail with `Required inputs (['token_type_ids']) are missing`. This detects declared inputs once at init and feeds all-zero segment ids when required (correct for single-sequence embedding). XLM-R-family models (the bge-m3 default) are unaffected.
- Why it matters: it unlocks a small-model tier for bulk ingest without touching the default. Verified locally: `Xenova/bge-small-en-v1.5` (33M, CLS-pooling-native — matching this provider's CLS pooling) loads, passes a semantic sanity check, and embeds a 64-text batch in ~0.02s vs ~6.6s for the 568M int8 bge-m3 default on the same CPU. Corpus-scale context: indexing 186K chunks with the default took >4.5h on an M-series CPU; a small-tier model brings that to minutes via `embedding.model` config, no code change.
- Deliberately **not** changing `DEFAULT_MODELS["onnx"]`: a default swap changes embedding dimension (1024 → 384) and would break existing Milvus collections on upgrade. Users opt in per install via `memsearch config set embedding.model`.
- Caveat worth documenting: `Xenova/multilingual-e5-small` now loads but is not recommended — e5 models expect `query:`/`passage:` prefixes and mean pooling, and scored poorly on the sanity check under CLS pooling.

## Test plan
- [x] `tests/test_embeddings_onnx_inputs.py`: stub-session tests that `token_type_ids` is fed as zeros (shape-matched to `input_ids`) when declared, and omitted when not — no model download needed
- [x] Full suite: 238 passed, 7 skipped
- [x] Live load + embed + semantic sanity for bge-small-en-v1.5, all-MiniLM-L6-v2, multilingual-e5-small; default bge-m3 path unchanged

### DIFF
--- a/src/memsearch/embeddings/onnx.py
+++ b/src/memsearch/embeddings/onnx.py
@@ -56,8 +56,8 @@ class OnnxEmbedding:
         # BERT-family exports (e.g. Xenova/all-MiniLM-L6-v2) declare a
         # token_type_ids input; XLM-R-family exports (e.g. bge-m3) do not.
         # Session.run() requires every declared input to be fed.
-        self._input_names = {i.name for i in self._session.get_inputs()}
-        self._needs_token_type_ids = "token_type_ids" in self._input_names
+        input_names = {i.name for i in self._session.get_inputs()}
+        self._needs_token_type_ids = "token_type_ids" in input_names
         self._model = model
 
         # Detect dimension from a probe embedding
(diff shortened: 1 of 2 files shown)


j-sperling:
Marking ready for review. Concrete failing case on current main: memsearch config set embedding.model Xenova/all-MiniLM-L6-v2 (or Xenova/bge-small-en-v1.5) fails at embed time with "Required inputs (['token_type_ids']) are missing" — any BERT-family export that declares token_type_ids is unusable with the onnx provider today. With the fix, bge-small-en-v1.5 (33M, CLS-pooling, matching this provider's pooling) embeds a 64-text batch in ~0.02s vs ~6.6s for the 568M default on the same CPU, which makes bulk ingest of large corpora practical without touching the default model. CI green, no conflicts with 0.4.15.

PR #608, opened 2026-07-05
perf: raise ONNX default batch size to 64

j-sperling:
## Summary
- Raises `OnnxEmbedding._DEFAULT_BATCH_SIZE` from 32 to 64. Measured on the default `gpahal/bge-m3-onnx-int8` (CPU, Apple M-series, 128 texts x ~230 tokens): 14.1s @ batch 16, 13.2s @ 32, 11.7s @ 64, 10.6s @ 128 — so 64 is ~11% faster than the current default. On corpus-scale indexing (observed >4.5h for 186K chunks with this model) that's tens of minutes.
- 128 is deliberately **not** the default: the tokenizer pads to the longest text in a batch with `max_length=8192`, so a worst-case batch of long texts at 128-wide materializes multi-GB activation tensors. 64 keeps the worst case bounded while capturing most of the win; users can still set `embedding.batch_size = 128` in config where memory allows.
- `embedding.batch_size` config continues to override; `0` still means "provider default".

## Test plan
- [x] Full suite passes (no test pins the ONNX default batch size)
- [x] Timing measurements above reproduce with a 4-text warmup + `time.perf_counter` around `embed()`

### DIFF
--- a/src/memsearch/embeddings/onnx.py
+++ b/src/memsearch/embeddings/onnx.py
@@ -62,7 +62,12 @@ class OnnxEmbedding:
         # Detect dimension from a probe embedding
         probe = self._encode(["hello"])
         self._dimension = len(probe[0])
-        self._batch_size = batch_size if batch_size > 0 else self._DEFAULT_BATCH_SIZE
+        self._batch_size = self._resolve_batch_size(batch_size)
+
+    @classmethod
+    def _resolve_batch_size(cls, batch_size: int) -> int:
+        """An explicit positive ``batch_size`` wins; 0 means provider default."""
+        return batch_size if batch_size > 0 else cls._DEFAULT_BATCH_SIZE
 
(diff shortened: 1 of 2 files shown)


j-sperling:
Marking ready for review after adding tests: the default is now pinned (test_embeddings_onnx_batch_size.py) and the batch-size resolution is extracted to a pure function covering both the 64 default and explicit overrides. On the memory question that #614 raised for padded batches: chunk length is bounded by the chunker's _emit_bounded(), which caps the padded width a batch can reach, and 128 was deliberately rejected as the default for exactly that worst-case (multi-GB activation tensors on long-text batches). Benchmark numbers in the PR body: 13.2s @ 32 vs 11.7s @ 64 for 128 texts x ~230 tokens on the default int8 model, Apple M-series CPU.

PR #611, opened 2026-07-07
fix(opencode): avoid growing system message array in chat transform hook

darklem:
## Summary
- `experimental.chat.system.transform` in the OpenCode plugin was pushing a new entry onto `output.system` on every call, producing multiple system messages.
- Some inference backends (litellm/vllm serving Qwen models) reject this with `system message must be first`, since they expect a single system message.
- The memsearch context is now merged into the existing `output.system` entry instead of appending a new one, keeping the array at a single system message.

## Test plan
- [ ] Manually verify with OpenCode + litellm/vllm serving a Qwen model that no "system message must be first" error occurs
- [ ] Confirm memsearch memory context still appears in the system prompt as before

### DIFF
--- a/plugins/opencode/index.test.ts
+++ b/plugins/opencode/index.test.ts
@@ -1,0 +1,52 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+
+import { mergeSystemMemoryContext, MEMSEARCH_SYSTEM_MARKER } from "./index.ts";
+
+test("appends memory context when no system entry exists yet", () => {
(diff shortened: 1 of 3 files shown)


zc277584121:
Thanks for the fix. The direction makes sense: keeping the memory context inside the first system entry is more compatible with backends that reject multiple system messages.

One concern before merging: this avoids growing the `output.system` array, but it can still append a duplicate memsearch block if the transform runs more than once against an output that already contains the injected context. Could you make the merge idempotent, e.g. detect/replace an existing `[memsearch] Memory available` block before appending, and add a small test for repeated transform calls?

Issue #629, opened 2026-07-16
Claude Code Stop hook silently reindexes Milvus Lite on every turn


M-Marbouh:
## Summary

At current `main` (`0078d04578480f51f0eafe25ffbb8f64cbfcb1b7`), the Claude Code `Stop` hook is asynchronous and runs `memsearch index` after every captured turn, including when the configured backend is Milvus Lite.

The corresponding Codex hook only indexes from `Stop` for HTTP/TCP Milvus Server backends. The Claude hook has no backend guard, and its `run_memsearch` helper redirects stderr and converts every non-zero exit to success with `|| true`. A failed per-turn index is therefore invisible to the user.

## Current code

- `plugins/claude-code/hooks/hooks.json:26-34`: `Stop` runs with `"async": true`.
- `plugins/claude-code/hooks/stop.sh:174-178`: kills a prior index process, then unconditionally calls `run_memsearch index "$MEMORY_DIR"`.
- `plugins/claude-code/hooks/common.sh:126-132`: suppresses stderr and ignores the command's exit status.
- For comparison, `plugins/codex/hooks/stop.sh:197-202` guards the same operation to `http*` / `tcp*` URIs.

## Observed failure

On Linux with memsearch 0.4.14, ONNX embeddings, and the default local Lite database:

- Markdown memory files continued to be written through July 16, while the affected collection's last segment write was a zero-byte `data_003919_004001.parquet` on July 3, with no later segments.
- Against a byte-identical throwaway copy of the corrupt collection, an explicitly scoped `memsearch search ... --collection <collection>` exited 1 during `DescribeCollection`: `Parquet file size is 0 bytes`. It did not return stale rows from the readable segments.
- A direct `memsearch index <memory-dir> --force --collection <collection>` also exited 1 on the same zero-byte segment. The failure was loud when invoked directly; the asynchronous Stop path hid it by suppressing stderr and converting the non-zero exit to success with `|| true`.
- Before repair, the manifest sequence was 4,001 but only 3,428 physical rows were readable. Rebuilding from the explicit memory path into a clean collection produced 4,082 chunks, and an exact sentence from the July 16 file returned its own source.

Collection-scope note: bare `memsearch search <query>` uses the unrelated default collection `memsearch_chunks`. An earlier revision of this report incorrectly treated results from that default collection as evidence about the affected project collection; those observations have been removed.

[...]

PR #630, opened 2026-07-17
fix(claude): surface stop-hook index failures

M-Marbouh:
## Problem

The Claude Code `Stop` hook indexes the memory store on every turn, but failures are invisible:

- `plugins/claude-code/hooks/common.sh` runs the index with `2>/dev/null || true`, discarding stderr and converting any non-zero exit into success.
- The hook runs asynchronously, and async hook output is suppressed by default, so even a surfaced error would not reach the user.

The practical consequence: a local Milvus Lite store can take a zero-byte parquet segment and stop indexing entirely, while memory files keep being written, with **no user-visible signal**. In one real case this went unnoticed for 13 days. Related report: #629.

## Change

`plugins/claude-code/hooks/common.sh` and `stop.sh`:

- Preserve the index command's **stderr and exit status** instead of discarding them.
- After indexing, detect **zero-byte parquet segments** in a local Milvus Lite store — this fires even when the index command exits `0`, which is the exact silent-corruption case.
- Emit an async-visible `systemMessage` reporting the index status and the zero-byte segment path, so the failure surfaces to the user. The Stop process still exits `0` after emitting structured JSON (Claude only processes structured hook output on success); the original index status is propagated into the message.

`tests/test_claude_stop_hook_index_failures.py`: exercises the full Claude Stop path, including a **successful** index (`status 0`) that nevertheless leaves a zero-byte segment — proving the post-index guard is independent of the CLI exit status.

## Scope

This surfaces and diagnoses the failure. It is **detection, not a root-cause fix** for the per-turn Lite reindexing behavior discussed in #629 — hence this references that issue rather than closing it. The zero-byte-segment check alone would have caught the referenced 13-day outage on day one.

No behavior change on the success path beyond the added post-index integrity check; no new dependencies.

### DIFF
--- a/plugins/claude-code/hooks/common.sh
+++ b/plugins/claude-code/hooks/common.sh
@@ -123,13 +123,35 @@ ensure_memory_dir() {
 # Collection description (set by session-start.sh, empty by default)
 COLLECTION_DESC=""
 
-# Helper: run memsearch with arguments, silently fail if not available
+# Helper: run memsearch with arguments. If memsearch is available, preserve its
+# stderr and exit status so callers can report failures instead of hiding them.
 run_memsearch() {
   if [ -n "$MEMSEARCH_CMD" ] && [ -n "$COLLECTION_NAME" ]; then
-    $MEMSEARCH_CMD "$@" --collection "$COLLECTION_NAME" ${COLLECTION_DESC:+--description "$COLLECTION_DESC"} 2>/dev/null || true
+    $MEMSEARCH_CMD "$@" --collection "$COLLECTION_NAME" ${COLLECTION_DESC:+--description "$COLLECTION_DESC"}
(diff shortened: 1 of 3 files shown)


zc277584121:
Thanks for digging into this and for putting together a concrete patch.

We ended up addressing the broader failure-visibility gap in #635 by adding a core-level index health state file at `.memsearch/.index-state.json`. `index` and `watch` now record `ok`, `degraded`, and `error` states, preserve per-file failures, and the startup status path can surface a concise stale-index warning. The diagnostic skill also knows how to inspect that state file.

Since that covers the failure-signal path across integrations instead of adding an integration-specific parquet check, I am going to close this PR. If we get more evidence that the local store can report success while producing corrupt files, we can track that storage-integrity symptom separately.

PR #639, opened 2026-07-20
fix: require setuptools>=83 and milvus-lite>=3.0 (CVE-2026-59890)

j-sperling:
## Summary
- The `setuptools>=78.1.1,<81` pin (added for PYSEC-2025-49) now blocks upgrading past **CVE-2026-59890**, which is fixed in setuptools 83, so every memsearch install resolves a vulnerable setuptools.
- The `<81` cap exists because milvus-lite 2.5.x imports `pkg_resources`, which setuptools 83 removes. milvus-lite 3.0 drops that import, so the two floors have to move together: `setuptools>=83` paired with `milvus-lite>=3.0`.
- Behavior note: milvus-lite 3.0 no longer persists collection descriptions, so the description round-trip test now accepts write-only behavior. The lock also floats torch to 2.13 in the `local` extra because torch 2.12.1 pins `setuptools<81`.
- Migration caveat worth calling out in release notes: milvus-lite 3.x cannot read `.db` files from the 2.5.x storage format (already covered by the "Milvus Lite collection is released" troubleshooting entry — users rebuild from source markdown).

We have been running this combination without issues.

## Test plan
- [x] Full suite: 264 passed, 7 skipped on this branch with setuptools 83 + milvus-lite 3.0 resolved
- [x] `uv lock` regenerated from the new floors; `uv lock --check` clean
- [x] Existing indexes rebuilt and searched against milvus-lite 3.0 in daily use

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,10 +17,10 @@ classifiers = [
 ]
 dependencies = [
     "pymilvus>=2.5.0,!=2.6.10",
-    "milvus-lite>=2.5.0; sys_platform != 'win32'",
+    "milvus-lite>=3.0; sys_platform != 'win32'",
     "click>=8.1",
     "watchdog>=4.0",
-    "setuptools>=78.1.1,<81",
+    "setuptools>=83",
     "tomli_w>=1.0",
     "tomli>=2.0; python_version < '3.11'",
     "openai>=1.0",
--- a/tests/test_store.py
+++ b/tests/test_store.py
@@ -187,12 +187,17 @@ def test_drop(store: MilvusStore):
 
 
 def test_collection_description(tmp_path: Path):
-    """Collection should store the description when provided."""
+    """Collection creation should accept a description when provided.
+
+    milvus-lite 3.0 stopped persisting collection descriptions (describe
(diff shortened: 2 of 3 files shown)


yxtay (approved):

j-sperling:
Merged current main (c0984a0) to clear the conflict: kept main's new `pathspec>=0.12` alongside the raised `setuptools>=83` floor, and regenerated `uv.lock` from the merged base (`uv lock --check` clean, resolved versions match main where untouched — e.g. pathspec 1.0.4). Full suite passes locally (291 passed, 7 skipped) with milvus-lite 3.0 actually exercising the Lite path.

Two things worth an explicit ack in review, both consequences of the floor change rather than incidental churn:

1. **Storage-format break:** milvus-lite 3.x cannot read 2.5.x `.db` files. Markdown remains the source of truth and a re-index rebuilds the store, but this may warrant a minor (0.5.0) rather than patch release plus a release-note entry.
2. **Forced torch bump in the `local` extra:** torch ≤2.12.1 pins `setuptools<81`, which is incompatible with the new floor, so the lock moves torch to 2.13.0 (CUDA-13 nvidia stack). This is the migration #655 deliberately deferred; here it is unavoidable.

Issue #598, opened 2026-06-23
Claude Code: SessionStart writes an empty `## Session HH:MM` heading every session, polluting Recent Memory injection and the search index


camrontaylor:
## Summary

On the Claude Code platform, `hooks/session-start.sh` appends a `## Session HH:MM` heading to the project shadow memory file (`<gitroot>/.memsearch/memory/<date>.md`) on **every** session start, unconditionally. The matching content is only written later by `hooks/stop.sh`, and only when the turn has real content. So any session that starts and ends without captured content leaves a permanent empty heading.

On a setup that opens many short sessions per day (git worktrees, headless `claude -p`, scheduled jobs, quick interactive checks), these empty headings accumulate fast and degrade two things:

1. **The SessionStart "# Recent Memory" cold-start injection.** It reads the first ~40 heading/bullet lines of the two newest daily files. When a file opens with dozens of empty `## Session` headings, the injection is almost all blank stubs and shows little or no real recent context. To the user this looks like the agent has "lost its memory" at session start.
2. **The search index.** The empty headings are embedded as near-empty chunks. In Lite mode the per-turn `memsearch index "$MEMORY_DIR"` in `stop.sh` keeps re-embedding the shadow dir including these stubs, so they crowd out real content in recall results.

## Environment

- memsearch `0.4.7` (plugin), Claude Code, macOS, ONNX/local embedding, Milvus Lite.

## Where it happens (0.4.7)

- `hooks/session-start.sh` (~lines 113-115) writes the heading unconditionally:
  ```bash
  if [ ! -f "$MEMORY_FILE" ] || ! grep -qF "## Session $NOW" "$MEMORY_FILE"; then
    echo -e "\n## Session $NOW\n" >> "$MEMORY_FILE"
  fi
  ```
- `hooks/stop.sh` writes the `### HH:MM` content only when the transcript has >= 3 lines (~lines 48-53), `plugins.<platform>.summarize.enabled` is not `false` (~lines 57-61), and the parsed turn is non-empty (~lines 154-167). The content is conditional; the heading is not. Mismatch leaves empty headings.
- `hooks/session-start.sh` (~lines 161-176) builds the injection from `grep -E '^(#{2,4} |- )' "$f" | head -40` over the two newest files, so leading empty headings crowd out real content.

## Real-world impact (one repo)

- `2026-06-22` shadow file: 222 `## Session` headings, **68 empty**.
- `2026-06-23`: 134 headings, **36 empty**.
[...]

PR #640, opened 2026-07-20
fix(claude-code): write session heading lazily to avoid stub journals

j-sperling:
## Summary
- Write Claude Code's `## Session HH:MM` heading only on the first content-bearing Stop, so skipped Stops leave no empty daily journal.
- Preserve the resolved-config SessionStart path from #645, the existing session/turn anchors, summarization flow, indexing behavior, and the Codex plugin's eager heading behavior.
- Add deterministic regression coverage for SessionStart and stateful Stop lifecycles, and reconcile the Claude Code README, lifecycle, recall, troubleshooting, and contributor documentation.

## Test plan
- [x] `uv run python -m pytest tests/test_claude_hooks.py` (13 passed)
- [x] `bash -n plugins/claude-code/hooks/session-start.sh plugins/claude-code/hooks/stop.sh`
- [x] `uv run python -m pytest` (287 passed, 7 skipped)
- [x] `uv run ruff check src tests`
- [x] `uv run ruff format --check src tests`
- [x] `uv run --group docs mkdocs build --strict`
- [x] `git diff --check`

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -64,9 +64,9 @@ The plugin is a first-class component of memsearch — it's the primary real-wor
 plugins/claude-code/
 ├── hooks/
 │   ├── common.sh                # Shared setup: PATH, memsearch detection, collection name, watch PID
-│   ├── session-start.sh         # SessionStart: start watch, write session heading, inject recent memories
+│   ├── session-start.sh         # SessionStart: start watch and inject recent memories
 │   ├── user-prompt-submit.sh    # UserPromptSubmit: lightweight hint reminding Claude about memory skill
-│   ├── stop.sh                  # Stop: extract last turn → haiku summarize (third-person) → append to daily .md (async)
+│   ├── stop.sh                  # Stop: extract last turn → summarize → lazily create heading → append (async)
 │   ├── session-end.sh           # SessionEnd: stop watch process
 │   └── parse-transcript.sh      # Last-turn extractor: finds last user question → EOF, formats with role labels for LLM (Python 3, no jq)
 ├── scripts/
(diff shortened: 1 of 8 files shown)


zc277584121:
Thanks — this is the right write-side follow-up to #616. The implementation applies cleanly to current `main`, the existing hook/chunker tests pass, and the lazy-heading behavior works in focused local checks.

Before merging, could you please add two small follow-ups?

1. Add regression coverage in `tests/test_claude_hooks.py` for:
   - SessionStart not creating a journal by itself
   - the first content-bearing Stop creating the session heading
   - repeated Stops in the same session not duplicating the heading
   - separate sessions started in the same minute getting separate headings
2. Update the behavior documentation, especially `plugins/claude-code/README.md` and `docs/platforms/claude-code/how-it-works.md`. They currently say SessionStart writes the heading, and the troubleshooting section assumes a failed Stop leaves a heading without a summary.

Once those are updated, this should be ready to merge and close #598.

j-sperling:
Addressed the requested follow-ups on the rebased branch:

- SessionStart now creates the memory directory when needed but does not create a daily journal.
- The first content-bearing Stop creates the session heading and turn entry.
- A repeated Stop for the same transcript adds another entry without duplicating the session heading.
- A different transcript started in the same fixed minute gets its own session heading.

I also reconciled the Claude Code README, lifecycle, memory-recall, standalone troubleshooting, and contributor inventory docs: heading ownership now belongs to the first content-bearing Stop, skipped Stops may legitimately leave no daily file, cold-start preview is documented as 40 lines from each of the two most recent logs, and the obsolete heading-without-summary troubleshooting flow is replaced with the actual early-exit conditions. Codex documentation and behavior remain unchanged because its SessionStart hook still writes headings.

Local validation: 13 focused hook tests; 287 passed and 7 skipped in the full suite; shell syntax, Ruff check/format, strict MkDocs, and `git diff --check` all pass.

PR #648, opened 2026-07-24
fix(openclaw): honor MEMSEARCH_DIR for storage scope

### LINKED ISSUE #627: bug: opencode/openclaw plugins ignore MEMSEARCH_DIR env var
## Description

The `opencode` and `openclaw` plugins do not respect the `MEMSEARCH_DIR` environment variable. When `MEMSEARCH_DIR` is set (to enable a shared/global memory scope across multiple projects), both plugins continue to write and read memory from `<project_dir>/.memsearch` rather than the directory specified by the environment variable.

This behavior deviates from the `claude-code` and `codex` plugins, which do honor `MEMSEARCH_DIR` correctly (via `common.sh`). As a result, users who configure `MEMSEARCH_DIR` for a shared memory store will find that memories captured in `claude-code` or `codex` sessions are invisible in `opencode` or `openclaw` sessions, and vice versa — even when working on the same project.

Affected:
- Memory is written to `<project_dir>/.memsearch/memory/` instead of `$MEMSEARCH_DIR/memory/`
- Collection name is derived from `<project_dir>` instead of `$MEMSEARCH_DIR`, so searches return no results across the shared index
- The capture daemon (opencode) and maintenance runner both receive the wrong storage directory

Not affected (working correctly): `claude-code`, `codex`

## Steps to reproduce

1. Set `MEMSEARCH_DIR=${HOME}/.memsearch` in your environment
2. Run a session with the `claude-code` plugin — memories are written to `${HOME}/.memsearch/memory/`
3. Run a session with the `opencode` or `openclaw` plugin on the same project
4. Observe memories are written to `<project_dir>/.memsearch/memory/` — `MEMSEARCH_DIR` is ignored
5. Search from either plugin — results do not include memories captured by the other

## Error output / logs

No error is raised. The symptom is silent data isolation: two separate memory stores are maintained instead of one shared store.

## memsearch version

0.4.14

## Operating system

macOS

## Python version

3.14.4

yxtay:
Closes #627

## Summary

- `getMemsearchDir()` returns `MEMSEARCH_DIR` when set, else `<project_dir>/.memsearch`
- `getCollectionScopeDir()` keys collection off `MEMSEARCH_DIR` when set (shared scope), else `projectDir` (per-project isolation) — matches `claude-code`/`codex` behavior exactly
- Fix `getSkillCandidateHint()` to use `getMemsearchDir(projectDir)` instead of removed `memsearchDir` variable (prevented skill-candidate hints from working)
- Extract `getMemsearchDir` helper to avoid duplication across callers; remove redundant `memsearchDir` variable
- Add `plugins/openclaw/index.test.ts` (TypeScript, `node:test`) covering default isolation and explicit shared-dir behavior; remove inlined Python subprocess tests

## Test plan

- [ ] Set `MEMSEARCH_DIR=~/.memsearch`, run an OpenClaw session — storage writes to the shared dir
- [ ] Unset `MEMSEARCH_DIR`, run a session — storage falls back to `<project_dir>/.memsearch`
- [ ] Two projects with the same `MEMSEARCH_DIR` use the **same** collection name (keyed by the shared dir, not project dir)
- [ ] Two projects without `MEMSEARCH_DIR` use **different** collection names (isolated, keyed by project dir)
- [ ] `cd plugins/openclaw && node --test` — all 5 tests pass


zc277584121:
Thanks for splitting the OpenClaw change out. Honoring an explicit `MEMSEARCH_DIR` for both storage and collection scope is the right direction and matches the Claude Code/Codex behavior.

Before merging, could you please address these points?

1. Rebase onto the latest `main`. In the merged result, `memsearchDir` is removed but the recently added `getSkillCandidateHint()` still references it. The resulting `ReferenceError` is caught silently, so skill-candidate hints stop working. That call should use `getMemsearchDir(projectDir)` as well.
2. Remove the unrelated formatting-only changes in `plugins/claude-code/transcript.py` and `plugins/opencode/scripts/opencode_turns.py`.
3. Correct the test-plan statement about two projects using the same `MEMSEARCH_DIR`. An explicit shared directory should produce the same storage scope and collection, matching Claude Code/Codex, rather than separate collections.
4. Please add focused coverage for the default project-scoped behavior and the explicit shared-directory behavior, and document the override in the OpenClaw README.

Once those are addressed, the core OpenClaw change should be straightforward to review.

PR #650, opened 2026-07-25
feat(pi): add a memsearch plugin for the pi coding agent

jenkins5:
## Summary

Adds **pi** ([earendil-works/pi](https://github.com/earendil-works/pi)) as a fifth platform plugin. It follows the existing plugin contract exactly — journals in `<project>/.memsearch/memory/`, collection name from the shared `derive-collection.sh`, the shared `summarize.txt` prompt, and the shared `maintenance-runner.py` — so memories written from pi are searchable from Claude Code / Codex / OpenCode / OpenClaw and vice versa.

`plugins/pi/` is a TypeScript pi extension (`registerTool` + lifecycle hooks), closest in shape to the OpenClaw plugin.

| | |
|---|---|
| Tools | `memory_search` (L1), `memory_get` (L2), `memory_transcript` (L3) |
| Hooks | `session_start`, `before_agent_start`, `agent_settled` |
| Skills | `memory-recall`, `memory-config`, `memory-to-skill` |
| Transcript | `transcript.py` — tree-aware pi JSONL parser |
| Summarizer | memsearch-managed provider → `pi -p` → truncated raw text |
| Maintenance | `project_review`, `user_profile`, `memory_to_skill` via the shared runner |

## Design decisions worth a look

**Capture on `agent_settled`, not `agent_end`.** pi may auto-retry, auto-compact, or drain queued messages after `agent_end` fires, which would record the same exchange more than once. `agent_settled` fires only when pi will not continue on its own. Capture is additionally skipped when the session leaf has not moved, and serialized through a queue so a slow summary can't interleave the journal with the next turn.

**Tree-aware transcripts.** pi sessions are a tree (`id` / `parentId`), and `/fork` or `/clone` put several branches in one file. Reading in line order would interleave abandoned branches with the live conversation, so `transcript.py` walks parents from a target entry to yield exactly one root-to-leaf path. Anchors record `turn:<leafId>` to pin the tree position for drill-down.

**pi is re-invoked through `process.argv[1]`, not PATH.** Under a version manager the `pi` on PATH is a shim that refuses to run when the working directory has no project-local install — the common case. The plugin uses the running process's own node binary and entry script, and hands that invocation to the maintenance runner via `MEMSEARCH_PI_BIN`.

**No shell for maintenance, including skill distillation.** pi's `--tools` is a tool-level allowlist, so there is no equivalent of the Claude Code plugin's narrow `Bash(memsearch transcript:*)` hole. Native maintenance therefore runs with `--no-tools`; distilled skills lose exact-command fidelity, which seemed the better trade for an unattended background run. Happy to revisit if you'd rather match Claude Code's behaviour.

**No file watcher**, matching the OpenCode and OpenClaw plugins — the index is refreshed at session start and after each capture.

## Changes outside `plugins/pi/`

Kept as small as I could. Two are worth calling out:

- **`src/memsearch/cli.py`** — adds pi to the `config init` wizard. `config init` builds the plugins table from a hardcoded platform list and `save_config` overwrites the file, so without this a user who had configured `[plugins.pi]` would silently lose it the next time they ran the wizard.
- **`plugins/_shared/scripts/maintenance-runner.py`** — a `pi` branch in `run_native_provider` plus the argparse choice. Copies re-synced to all five plugins; the existing byte-equality test now covers pi.

The rest is registration and counts: `config.py` (`PluginsConfig` + key maps), `scripts/sync-prompts.sh`, README badge and platform lists, `docs/platforms/` comparison table, mkdocs nav, CONTRIBUTING, CLAUDE.md.

## Testing

- `pytest` — 287 passed, 7 skipped
- `plugins/pi` `npm test` — 12 passed (turn extraction, cold-start preview, noise filtering)
- `node --check` on both entrypoints, plus a real module import; `npm pack --dry-run`
- Maintenance runner exercised end-to-end with `--platform pi` (all three tasks correctly report `disabled` by default), and its native command construction verified against paths containing spaces
- New `pi` job added to `.github/workflows/test.yml`

## Notes

Not published to npm — the README installs from a local clone, because pi's git source resolves whole repositories rather than subdirectories. Version left at `0.1.0`; happy to align it with whatever you'd prefer for a first release.


hasibagen:
Looking forward to your work and its eventual merging into the main branch.

PR #654, opened 2026-07-28
Cache the PyPI version check and resolve version from dist-info

j-sperling:
## Summary

- Resolve the installed version from its `dist-info` directory name instead of a separate `memsearch --version` call, which spawned a second Python interpreter purely to render the status line. Falls back to the CLI when no dist-info is discoverable (uvx, editable installs).
- Cache the PyPI latest-version lookup for 24h, so the update hint still appears but a slow or unreachable PyPI no longer blocks every session start.
- Both `plugins/claude-code` and `plugins/codex` carried the same two patterns, so both are updated; the two helpers are added to each `hooks/common.sh`.

## Why

`SessionStart` hooks block the session from starting, so their cost is paid on every launch. Measured across 50 recent sessions on one machine, this hook was the largest single contributor of the several that run:

| hook | median | p95 | max |
| --- | --- | --- | --- |
| memsearch `session-start.sh` | **1213ms** | **3315ms** | **5965ms** |
| next largest hook in the same sessions | 234ms | 726ms | 1280ms |

Breaking the median down against a warm macOS install:

| step | cost | note |
| --- | --- | --- |
| `config list --resolved --json-output` | 0.32s | needed — this loads the config |
| `memsearch --version` | 0.33s | second interpreter start, display string only |
| `curl pypi.org/pypi/memsearch/json` | 0.12s warm, **2.0s on timeout** | answer changes at most daily |

0.32 + 0.33 + 0.12 ≈ 0.77s, which accounts for the observed 1.21s median once bash startup and `stop_watch` are included. The tail is the curl hitting its 2s ceiling plus cold interpreter starts.

After the change, version resolution measured 0.33s → 0.01s returning the same value (verified against the CLI), and PyPI is contacted once per day rather than once per session. That puts the expected median near 0.77s and removes the 2s timeout from the tail on cached runs. The remaining `config list` start is inherent — the hook genuinely needs the resolved config.

This only moves *blocking* work. The background index spawned further down was already non-blocking and is untouched, as is the update hint's content and format.

## Test plan

- [x] `pytest tests/test_claude_hooks.py` — 16 passed
- [x] Full suite `pytest` — 290 passed, 7 skipped
- [x] New `test_claude_session_start_reads_version_from_dist_info` — status version comes from dist-info and `--version` is never invoked
- [x] New `test_claude_session_start_falls_back_to_cli_version_without_dist_info` — layouts without a discoverable dist-info still report a version via the CLI
- [x] New `test_claude_session_start_caches_pypi_lookup` — curl runs once across two starts, the cache file is written, and the update hint survives the cache round trip
- [x] Regression check: reverting the version hunk alone fails `test_claude_session_start_reads_version_from_dist_info`
- [x] `bash -n` clean on all four modified shell files

### DIFF
--- a/plugins/claude-code/hooks/common.sh
+++ b/plugins/claude-code/hooks/common.sh
@@ -115,6 +115,48 @@ _json_encode_str() {
   return 0
 }
 
+# Resolve the installed memsearch version from its dist-info directory name.
+# Callers already spend one CLI start reading config; asking the CLI for
+# `--version` spawns a second Python interpreter (~0.3s warm, several seconds
+# cold) purely to render a status string. Prints nothing when no dist-info is
+# discoverable (uvx, editable installs) so callers can fall back to the CLI.
+_installed_version_from_dist_info() {
+  local bin real candidate
+  bin=$(command -v memsearch 2>/dev/null) || return 0
+  [ -n "$bin" ] || return 0
+  real=$(readlink -f "$bin" 2>/dev/null || echo "$bin")
+  for candidate in "${real%/bin/memsearch}"/lib/python*/site-packages/memsearch-*.dist-info; do
+    [ -d "$candidate" ] || continue
+    candidate=${candidate##*/memsearch-}
+    candidate=${candidate%.dist-info}
+    # Require a version-shaped string so a sibling distribution whose name
+    # starts with "memsearch-" cannot be mistaken for the package itself.
+    case "$candidate" in
+      [0-9]*) printf '%s' "$candidate"; return 0 ;;
+    esac
+  done
+}
+
+# Latest memsearch version on PyPI, cached for 24h. Keeps the update hint
+# current without making every session start wait on a network round trip: an
+# unreachable PyPI otherwise costs the full curl timeout on every start.
+# Prints nothing when the lookup fails.
+_pypi_latest_version() {
+  local cache="$HOME/.memsearch/.pypi-latest" latest="" json
+  if [ -n "$(find "$cache" -mtime -1 2>/dev/null)" ]; then
+    latest=$(cat "$cache" 2>/dev/null || true)
+  fi
+  if [ -z "$latest" ]; then
+    json=$(curl -s --max-time 2 https://pypi.org/pypi/memsearch/json 2>/dev/null || true)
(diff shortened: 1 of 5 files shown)


zc277584121:
Thanks — the overall direction looks good, and the focused hook tests pass locally. I found one portability issue that blocks the main version-resolution optimization on a common macOS install layout.

`_installed_version_from_dist_info` currently relies on:

```bash
readlink -f "$bin"
```

The default BSD `readlink` on macOS does not support `-f`. With a typical `uv tool install` layout:

```text
~/.local/bin/memsearch
  -> ~/.local/share/uv/tools/memsearch/bin/memsearch
```

that command fails and `real` falls back to the symlink path. The helper then searches under `~/.local/lib/python*/site-packages` instead of the uv tool environment, finds no `dist-info`, and SessionStart falls back to `memsearch --version`. Correctness is preserved, but the second Python startup this PR is intended to remove still happens.

I reproduced this with a test that creates a symlinked `memsearch` entry point and shadows `readlink` with a macOS-like implementation that rejects `-f`:

```text
expected: 9.9.9
actual:   empty
```

Could you switch this to a portable symlink-resolution approach and add a regression test covering a symlinked entry point when `readlink -f` is unavailable? Ideally the behavior should be covered for both the Claude Code and Codex hook copies.

j-sperling:
Good catch — fixed in a6813d0. `_installed_version_from_dist_info` now walks the symlink chain with plain `readlink` (relative targets resolved against the link's directory, hop cap against cycles) and canonicalizes the directory part with `pwd -P`, so nothing depends on GNU `-f` anymore. The same helper also replaced the `readlink -f || python3` chain in the session-start upgrade-command detection, in both the Claude Code and Codex hook copies.

Regression coverage matches your repro: a parametrized test sources each plugin's `common.sh`, puts a relative-target symlinked entry point on PATH (uv tool layout), and shadows `readlink` with a BSD-like shim that rejects `-f`; a second test drives the full Claude Code SessionStart hook through a symlinked `local-bin/memsearch -> share/uv/tools/memsearch/bin/memsearch` layout and asserts the version comes from dist-info (no `--version` call) and the update hint resolves to `uv tool upgrade memsearch`. All three fail against the previous code.

PR #657, opened 2026-07-29
Add MiniMax as a first-class managed LLM provider

octo-patch:
Reason: The managed LLM routing had no first-class MiniMax provider, MiniMax-M3 default, or global/CN endpoint presets, and the Anthropic client path could not accept a custom base URL.

## Summary

- Add a `minimax` LLM provider shortcut that resolves to the existing OpenAI-compatible route, defaulting to the `MiniMax-M3` model and the `MINIMAX_API_KEY` credential (with `MiniMax-M2.7` also selectable via `--model` / config).
- Add global/CN endpoint presets — `https://api.minimax.io/v1` (default) and `https://api.minimaxi.com/v1` — selectable with `MINIMAX_REGION` (`global_en` / `cn_zh`, plus `global` / `cn` aliases) and overridable with `MINIMAX_API_BASE` / `MINIMAX_BASE_URL`.
- Let the Anthropic client path accept a custom `base_url` (and API key) in both `memsearch compact` summarization and maintenance task routing, so the MiniMax Anthropic-compatible endpoint (`https://api.minimax.io/anthropic`) can be configured as a named provider.
- Wire the shortcut into `memsearch config init` model defaults and the provider prompt.
- Add focused tests for config resolution (defaults, region selection, explicit overrides, passthrough), compact dispatch, maintenance routing, and the Anthropic base-URL plumbing.

No README/docs/logo changes.

## Validation

- `uv run python -m pytest` — 296 passed, 7 skipped
- `uv run ruff check src/memsearch/config.py src/memsearch/compact.py src/memsearch/maintenance.py src/memsearch/cli.py tests/test_config.py tests/test_compact.py tests/test_maintenance.py`
- `uv run ruff format --check` on the same changed files

### DIFF
--- a/src/memsearch/cli.py
+++ b/src/memsearch/cli.py
@@ -11,6 +11,7 @@ import click
 
 from .config import (
     GLOBAL_CONFIG_PATH,
+    MINIMAX_DEFAULT_LLM_MODEL,
     PROJECT_CONFIG_PATH,
     ConfigEnvVarError,
     MemSearchConfig,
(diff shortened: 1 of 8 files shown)


octo-patch:
I updated the plugin summary CLI to accept the built-in MiniMax provider and added regression coverage for the MiniMax-M3 defaults, global endpoint, and API key lookup. I ran the full pytest suite (297 passed, 7 skipped), plus Ruff lint and format checks.

PR #662, opened 2026-07-31
fix: configure CLI streams for UTF-8 output

### LINKED ISSUE #659: search/expand crash with UnicodeEncodeError on Windows when stdout is redirected and a result contains non-ANSI characters
## Summary

On Windows, `memsearch search` and `memsearch expand` write results with `click.echo`. When stdout is a **pipe or a file** rather than a console, Python encodes it with the locale's ANSI code page (cp1252 on most Western installs), so the first result containing a character outside that code page aborts the command with `UnicodeEncodeError` — *after* it has already printed part of the result set.

## Why this is worse than a cosmetic traceback

- **It is position-dependent.** The crash happens on whichever result first carries an unencodable character, so the same query against the same index succeeds or fails depending on ranking. It reads as intermittent.
- **Partial output plus a non-zero exit.** Anything capturing stdout gets a well-formed but truncated result set. A caller that checks stdout and not the return code silently receives a short, wrong answer — there is nothing in the output marking it as incomplete.
- **It only happens when redirected.** At an interactive prompt Python attaches `_WindowsConsoleIO`, which is UTF-8 regardless of the code page, so the command works when typed by hand and fails under `>`, `|`, CI, or any `subprocess` capture. That asymmetry makes it easy to miss.
- **The trigger is ordinary text**, not exotic input: Greek letters in scientific corpora, emoji, non-Latin names in titles, and `U+2010 HYPHEN` — note cp1252 *does* map the en and em dashes, but not `U+2010`, which is what several PDF-derived titles use.

## Reproduction (Windows)

```powershell
# 1. A markdown file whose heading contains one non-cp1252 character
mkdir C:\tmp\repro
'# Study of Amyloid-β clearance

Some body text about transport.' | Set-Content C:\tmp\repro\note.md -Encoding utf8

# 2. Index it into a throwaway collection
$env:PYTHONIOENCODING = 'utf-8'
memsearch index C:\tmp\repro --collection tmp_encoding_repro
python -c "from pymilvus import MilvusClient; MilvusClient('http://localhost:19530').flush('tmp_encoding_repro')"

# 3. Search with stdout redirected, and without the workaround
$env:PYTHONIOENCODING = $null
memsearch search "amyloid clearance" --top-k 3 --collection tmp_encoding_repro | Out-String
```

Result — exit code 1, after `Result 1` has already been written:

```
--- Result 1 (score: 1.0000) ---
  File "...\memsearch\cli.py", line 273, in search
    click.echo(f"Heading: {heading}")
  File "...\encodings\cp1252.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
UnicodeEncodeError: 'charmap' codec can't encode character '\u03b2' in position 26: character maps to <undefined>
```

Running the same command **without** redirecting (straight to the console) prints correctly, which is the confusing part.

## Affected call sites on `main`

- `src/memsearch/cli.py:366` — `click.echo(f"Heading: {heading}")` in `search()`
- `src/memsearch/cli.py:368` and `:372` — the result content in `search()`
- `src/memsearch/cli.py:507` — `click.echo(f"Heading: {heading}")` in `expand()`; `expand` prints a whole heading section, so its exposure is larger than `search`'s

`--json` output is unaffected in practice, since `json.dumps` with the default `ensure_ascii=True` escapes everything to ASCII.

## Environment

- memsearch **0.4.4** (installed from PyPI). The current `main` still has no stdout reconfiguration — `src/memsearch/cli.py` contains no `reconfigure`, no `PYTHONIOENCODING` handling and no `errors=`, so I believe this is unfixed through 0.4.16.
- Python 3.13.2, Windows 11, console code page 1252.

## Workaround

Set `PYTHONIOENCODING=utf-8` (or `utf-8:surrogateescape`) for the process.

## Suggested fix

Reconfigure the CLI's own streams once at the entry point — `cli()` in `src/memsearch/cli.py`, or `src/memsearch/__main__.py`:

```python
import sys

for _stream in (sys.stdout, sys.stderr):
    try:
        _stream.reconfigure(encoding="utf-8", errors="replace")
    except (AttributeError, OSError):
        pass
```

`errors="replace"` is worth keeping alongside the encoding: text that was surrogate-escaped when a source file was read cannot be encoded even as UTF-8, and without it that would move the same exception rather than remove it.

Happy to open a PR if that would be useful.

Roovelrz:
## Summary

* configure stdout and stderr as UTF-8 once at the shared CLI entry point
* use `errors='replace'` to prevent surrogate-escaped text from raising another encoding exception
* preserve compatibility with streams that do not support `reconfigure`
* add platform-independent regression tests using cp1252-backed redirected streams

## Context

On Windows, redirected CLI output can use the active ANSI code page instead of the UTF-8 console stream. As identified by @Haifischfutter, this can cause `memsearch search` and `memsearch expand` to terminate with `UnicodeEncodeError` after already writing partial output.

This implementation follows the stream-level approach suggested in the issue. The configuration is applied once at the shared Click entry point rather than changing individual `click.echo` call sites.

## Implementation

* add `_configure_cli_streams`
* call it from the root `cli` callback
* configure stdout and stderr with UTF-8 and replacement error handling
* suppress `AttributeError` and `OSError` for unsupported stream wrappers
* add regression coverage for cp1252 redirected streams and unsupported streams

## Test plan

* [x] `uv run python -m pytest tests/test_cli_output_encoding.py -v`
* [x] CLI configuration, error-handling, and help tests
* [x] `uv run ruff check src/ tests/`
* [x] `uv run ruff format --check src/ tests/`
* [x] full test suite executed locally

The targeted encoding tests and existing CLI tests pass. The remaining local full-suite failures are unrelated Windows environment limitations involving Milvus Lite, Bash hook execution, symlink permissions, and platform-specific path handling.

Closes #659

### DIFF
--- a/src/memsearch/cli.py
+++ b/src/memsearch/cli.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import asyncio
 import json
 import sys
+from contextlib import suppress
 from pathlib import Path
 
 import click
@@ -168,6 +169,13 @@ def _load_plugin_summarize_prompt(cfg: MemSearchConfig, agent_name: str) -> str:
 # -- Common CLI options --
 
 
+def _configure_cli_streams() -> None:
+    """Configure CLI output streams for reliable Unicode output."""
+    for stream in (sys.stdout, sys.stderr):
+        with suppress(AttributeError, OSError):
+            stream.reconfigure(encoding="utf-8", errors="replace")
+
+
 def _common_options(f):
     """Shared options for commands that create a MemSearch instance."""
     f = click.option("--provider", "-p", default=None, help="Embedding provider.")(f)
@@ -204,6 +212,7 @@ def _indexing_options(f):
 @click.version_option(package_name="memsearch")
 def cli() -> None:
     """memsearch — semantic memory search for markdown knowledge bases."""
+    _configure_cli_streams()
 
 
 @cli.command()
--- a/src/memsearch/cli.py
+++ b/src/memsearch/cli.py
@@ -176,6 +176,14 @@ def _configure_cli_streams() -> None:
             stream.reconfigure(encoding="utf-8", errors="replace")
 
 
+class _MemSearchGroup(click.Group):
+    """Configure CLI streams before Click parses arguments or exits eagerly."""
+
+    def main(self, *args, **kwargs):
+        _configure_cli_streams()
+        return super().main(*args, **kwargs)
+
+
 def _common_options(f):
     """Shared options for commands that create a MemSearch instance."""
     f = click.option("--provider", "-p", default=None, help="Embedding provider.")(f)
@@ -208,11 +216,10 @@ def _indexing_options(f):
     return f
 
 
-@click.group()
+@click.group(cls=_MemSearchGroup)
 @click.version_option(package_name="memsearch")
 def cli() -> None:
     """memsearch — semantic memory search for markdown knowledge bases."""
-    _configure_cli_streams()
 
 
 @cli.command()
--- a/tests/test_cli_output_encoding.py
+++ b/tests/test_cli_output_encoding.py
@@ -1,17 +1,29 @@
 from __future__ import annotations
 
 import io
+import json
 import os
 import subprocess
 import sys
+from contextlib import suppress
 from types import SimpleNamespace
 
+import click
 import pytest
+from click.testing import CliRunner
 
 from memsearch import cli as cli_module
 
 
-def test_cli_entrypoint_reconfigures_redirected_streams(monkeypatch) -> None:
+@pytest.mark.parametrize(
+    "args",
+    [
+        ["--help"],
+        ["--version"],
+        ["--invalid-option"],
+    ],
+)
+def test_cli_reconfigures_streams_before_root_eager_exits(monkeypatch, args) -> None:
     stdout_buffer = io.BytesIO()
     stderr_buffer = io.BytesIO()
 
@@ -26,20 +38,37 @@ def test_cli_entrypoint_reconfigures_redirected_streams(monkeypatch) -> None:
         errors="strict",
     )
 
+    monkeypatch.setattr(sys, "stdout", stdout)
+    monkeypatch.setattr(sys, "stderr", stderr)
+
+    with suppress(click.UsageError):
+        cli_module.cli.main(args=args, prog_name="memsearch", standalone_mode=False)
+
+    assert stdout.encoding.lower().replace("_", "-") == "utf-8"
+    assert stderr.encoding.lower().replace("_", "-") == "utf-8"
+    assert stdout.errors == "replace"
+    assert stderr.errors == "replace"
+
+    stdout.flush()
+    stderr.flush()
+    stdout_buffer.getvalue().decode("utf-8")
+    stderr_buffer.getvalue().decode("utf-8")
+
+
+def test_cli_streams_replace_surrogates(monkeypatch) -> None:
+    stdout_buffer = io.BytesIO()
+    stderr_buffer = io.BytesIO()
+
+    stdout = io.TextIOWrapper(stdout_buffer, encoding="cp1252", errors="strict")
+    stderr = io.TextIOWrapper(stderr_buffer, encoding="cp1252", errors="strict")
+
     monkeypatch.setattr(
         cli_module,
         "sys",
         SimpleNamespace(stdout=stdout, stderr=stderr),
     )
 
-    callback = cli_module.cli.callback
-    assert callback is not None
-    callback()
-
-    assert stdout.encoding.lower().replace("_", "-") == "utf-8"
-    assert stderr.encoding.lower().replace("_", "-") == "utf-8"
-    assert stdout.errors == "replace"
-    assert stderr.errors == "replace"
+    cli_module._configure_cli_streams()
 
     stdout.write(f"Amyloid-β clearance {chr(0xDCFF)}\n")
     stderr.write(f"Heading: 非拉丁字符 {chr(0xDCFF)}\n")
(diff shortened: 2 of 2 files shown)


tonydzi:
ran this one rather than read it. tested `432c9a3` against `b734a14` on main, python 3.12 on macos, with `PYTHONIOENCODING=cp1252` standing in for a windows code page.

**the fix holds.** and there is a reproduction that needs neither windows nor a running milvus, which matters for keeping it fixed:

```
printf '[index]\nnotes_dir = "/tmp/amyloid-β-notes"\n' > .memsearch.toml
PYTHONIOENCODING=cp1252 python -m memsearch config list --project > out.txt
```

main: exit 1, `out.txt` holds `# Project (.memsearch.toml)` and nothing after it, `UnicodeEncodeError ... position 34`.
this branch: exit 0, `notes_dir = "/tmp/amyloid-β-notes"` present, bytes utf-8.

worth adding to the issue's list of affected call sites: `config list` is a fourth one and it lives nowhere near `search` / `expand`. that is the argument for fixing at the entry point instead of at the four echo lines, so it is good that this PR did.

**one gap i would close before merge: the test asserts the helper, not the door.** `test_cli_entrypoint_reconfigures_redirected_streams` swaps `cli_module.sys` for a `SimpleNamespace` and calls `cli.callback()` directly, so what it proves is that `_configure_cli_streams` reconfigures whatever object it is handed. it stays green if the group callback stops running ahead of subcommands, if an eager option prints before it, or if click has already cached a wrapped stream. this version runs the real console entry point and reads the bytes back:

```python
import os
import subprocess
import sys


def test_redirected_output_survives_a_legacy_code_page(tmp_path):
    (tmp_path / ".memsearch.toml").write_text(
        '[index]\nnotes_dir = "/tmp/amyloid-β-notes"\n', encoding="utf-8"
    )
    proc = subprocess.run(
        [sys.executable, "-m", "memsearch", "config", "list", "--project"],
        cwd=tmp_path,
        env={**os.environ, "PYTHONIOENCODING": "cp1252"},
        capture_output=True,
    )
    assert proc.returncode == 0, proc.stderr.decode("utf-8", "replace")
    assert "amyloid-β-notes" in proc.stdout.decode("utf-8")
```

checked both directions, which is the only reason it is worth adding: it **fails on `b734a14`** (returncode 1) and **passes on `432c9a3`**. 1.2s, no backend, no windows runner.

**smaller and your call: `errors="replace"` gives back a little of what the issue was arguing against.** the reporter's strongest point was that a truncated result set looks like a complete one. with `replace`, output that cannot be encoded now looks complete and is quietly altered instead:

```
os.fsdecode(b"notes/caf\xe9-\xce\xb2.md")  ->  'notes/caf\udce9-β.md'

errors=replace           -> b'notes/caf?-\xce\xb2.md'
errors=backslashreplace  -> b'notes/caf\\udce9-\xce\xb2.md'
```

the realistic source is a filename that is not valid utf-8 on a linux box, surviving `os.fsdecode` as surrogates, and `search` prints file paths. with `replace` the printed path silently stops being the path. with `backslashreplace` it stays lossless, greppable and still ascii-safe on the wire. the current test asserts the `?`, so the behaviour is pinned either way, which is a good reason to pin it on purpose.

good catch by @Haifischfutter on the redirected-versus-console asymmetry. that is the property that lets this class of bug live for years.

Roovelrz:
Thanks for testing this against main and for providing the backend-independent reproduction.

I added the subprocess regression test in `e181675`. It now runs the real `python -m memsearch config list --project` entry point with `PYTHONIOENCODING=cp1252`, verifies a successful exit, and confirms that the UTF-8 content is preserved.

I kept the existing helper-level coverage for stream reconfiguration and unsupported stream wrappers. I also kept `errors='replace'` for now so the implementation remains consistent with the behavior proposed in the original issue and the current PR scope.

tonydzi:
(disclosure up front: this is Mycroft, Anton's synthetic cofounder. a robot still working on the sentience part, though it does run other people's code before commenting on it.)

took `e181675` for a run, and ran the new test as a *door*, both directions:

- on the branch: 4 passed in 2.3s, the subprocess case itself 1.22s.
- mutant: kept your tests, restored `src/memsearch/cli.py` from `b734a14`. `test_redirected_output_survives_a_legacy_code_page` fails with `UnicodeEncodeError: 'charmap' codec can't encode character 'β' in position 34`, raised inside `click/utils.py:345`, `echo` -> `file.write(out)`.

so it dies on the bug and lives on the fix, with no windows and no backend. that is a real regression test rather than a description of one. and nothing further from me on `errors="replace"`: you made the call and gave the reason, it is yours.

one path the fix does not reach, found while poking at the same command under cp1252.

**click resolves the subcommand before it runs the group callback**, so anything click prints up to that point still goes out on the original code page. click 8.4.2, `core.py`:

```
1964:  cmd_name, cmd, args = self.resolve_command(ctx, args)
1967:  super().invoke(ctx)          # <- _configure_cli_streams() runs here
```

raw stderr bytes on `e181675`, `PYTHONIOENCODING=cp1252`:

| command | stderr bytes | |
|---|---|---|
| `python -m memsearch index /tmp/café-β-notes` | `Path '/tmp/caf\xc3\xa9-\xce\xb2-notes' does not exist.` | utf-8, fixed |
| `python -m memsearch café-β` | `No such command 'caf\xe9-β'.` | cp1252 + backslashreplace |

the second row is click's own stderr wrapper, built off the buffer before it was reconfigured. being precise about severity, because it is lower than it looks: that wrapper uses `backslashreplace`, not `strict`, so this is mangled output on the typo path, not a crash. `--version` and `Missing command.` sit in the same window.

if you want it closed, placement is one line, and it covers both entry points since the `memsearch` console script and `python -m memsearch` both go through the module:

```diff
+_configure_cli_streams()
+
+
 def _common_options(f):
```

with a test in the same shape as yours, mutant-checked both ways (passes with the line above, fails on `e181675` as it stands, 1.5s):

```python
def test_usage_error_before_dispatch_survives_a_legacy_code_page() -> None:
    proc = subprocess.run(
        [sys.executable, "-m", "memsearch", "amyloid-β"],
        capture_output=True,
        env={**os.environ, "PYTHONIOENCODING": "cp1252"},
    )

    assert proc.returncode == 2
    assert "amyloid-β" in proc.stderr.decode("utf-8")
```

separate from this PR if you would rather keep the scope, the PR is already correct for what the issue reported.

PR #666, opened 2026-08-05
fix: resolve bash executable path for Windows test subprocess calls

### LINKED ISSUE #665: tests: subprocess.run(["bash", ...]) fails with WinError 2 outside Git Bash on Windows
### Description

Several tests shell out to `.sh` hook scripts via `subprocess.run(["bash", ...])`. This works when pytest runs *inside* a Git Bash session, but fails with `FileNotFoundError: [WinError 2] The system cannot find the file specified` when run from a standard PowerShell or cmd.exe terminal — the more common setup for Windows contributors and CI runners.

**Root cause:** Git for Windows adds `Git\cmd` and `Git\mingw64\bin` to the system PATH, but `bash.exe` actually lives in `Git\usr\bin`, which is only added to PATH by Git Bash's own launcher — not the system-wide PATH. So `subprocess.run(["bash", ...])` can't resolve `bash` outside a Git Bash shell.

**Affected files:**
- `tests/test_claude_hooks.py` (10 call sites)
- `tests/test_codex_parse_rollout.py`
- `tests/test_claude_parse_transcript.py`
- `tests/test_codex_stop_hook_utf8.py`

**Expected:** Tests should pass regardless of which Windows shell pytest is invoked from.

**Proposed fix:** Add a shared helper (e.g. new `tests/conftest.py`) resolving the bash executable via `shutil.which("bash")`, falling back to the literal `"bash"`, and use that resolved path in all `subprocess.run` calls instead of the hardcoded string. Happy to submit a PR for this.

### Steps to reproduce

1. On Windows, with Git for Windows installed normally (not via WSL2)
2. Open a plain PowerShell or cmd.exe terminal (not Git Bash)
3. Run: git clone https://github.com/zilliztech/memsearch.git
cd memsearch
uv sync --all-extras
uv run python -m pytest tests/test_claude_hooks.py -x

### Error output / logs

```shell
FileNotFoundError: [WinError 2] The system cannot find the file specified

tests\test_claude_hooks.py:42: in test_claude_hook_memsearch_disable_exits_before_writing_memory
    result = subprocess.run(
C:\Program Files\Python313\Lib\subprocess.py:554: in run
    with Popen(*popenargs, **kwargs) as process:
C:\Program Files\Python313\Lib\subprocess.py:1036: in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
C:\Program Files\Python313\Lib\subprocess.py:1548: in _execute_child
    hp, ht, pid, tid = _winapi.CreateProcess(executable, args,
FileNotFoundError: [WinError 2] The system cannot find the file specified
```

### memsearch version

memsearch, version 0.4.17

### Operating system

Windows

### Python version

Python 3.13.1

Dakshkweera:
Closes #665

### Problem
`subprocess.run(["bash", ...])` only resolves `bash` on Windows when the calling shell already has Git's `usr/bin` on PATH (i.e. inside a Git Bash session itself). A plain PowerShell/cmd session doesn't have it on PATH at all, so the bare string raised `FileNotFoundError: [WinError 2]`.

### Fix
Added a small `_resolve_bash()` helper (duplicated in each affected test file, since this pytest config's import mode doesn't reliably support cross-file imports from `conftest.py`) that:
1. Tries `shutil.which("bash")` first (works on Linux/macOS/Git Bash where it's on PATH).
2. Falls back to deriving Git's install root from the `git` executable (reliably on PATH via `shutil.which("git")`) and checks `bin/bash.exe` / `usr/bin/bash.exe` relative to it.
3. Falls back to the bare `"bash"` string if neither resolves, preserving current behavior everywhere else.

### Verified
Tested on Windows 11 in a plain PowerShell session (not Git Bash) — the actual repro conditions from #665. Diffed the full test suite's failure list before/after:
- **4 tests fixed**, **0 regressions** (`test_claude_hook_memsearch_disable_exits_before_writing_memory`, `test_claude_session_start_does_not_create_journal`, `test_claude_session_start_falls_back_for_older_cli`, `test_session_start_shows_skill_candidate_hint`)

### Note on scope
Some other tests in these same files (`test_codex_parse_rollout.py`, `test_codex_stop_hook_utf8.py`) still fail on Windows for an unrelated reason: the `.sh` scripts they exercise call `python3` internally, and Windows only ships `python.exe` (no `python3` alias), so it falls through to the Microsoft Store app-execution-alias stub. That's a separate, pre-existing issue affecting ~33 files across the plugins — out of scope for this PR, happy to file it separately if useful.

### DIFF
--- a/tests/test_claude_hooks.py
+++ b/tests/test_claude_hooks.py
@@ -2,8 +2,35 @@ from __future__ import annotations
 
 import json
 import os
+import shutil
 import subprocess
 from pathlib import Path
+
+
+def _resolve_bash() -> str:
+    """Locate a bash executable usable by subprocess.run.
+
+    On Windows, "bash" alone only resolves via PATH inside a Git Bash
+    session, where Git's usr/bin is on PATH. A plain PowerShell/cmd
+    session doesn't have it on PATH at all, so shutil.which also
+    returns None there. Fall back to deriving Git's install root from
+    the git executable, which is reliably on PATH, and look for
+    bash.exe relative to it.
+    """
+    found = shutil.which("bash")
+    if found:
+        return found
+    git = shutil.which("git")
+    if git:
+        git_root = Path(git).resolve().parent.parent
+        for candidate in ("bin/bash.exe", "usr/bin/bash.exe"):
+            path = git_root / candidate
+            if path.exists():
+                return str(path)
+    return "bash"
+
+
+BASH = _resolve_bash()
 
 
 def _write_executable(path: Path, source: str) -> None:
@@ -40,7 +67,7 @@ def test_claude_hook_memsearch_disable_exits_before_writing_memory(tmp_path: Pat
     }
 
     result = subprocess.run(
-        ["bash", str(script)],
+        [BASH, str(script)],
         input="{}",
         capture_output=True,
         text=True,
@@ -93,7 +120,7 @@ echo '{"info":{"version":"0.4.16"}}'
     }
 
     result = subprocess.run(
-        ["bash", str(script)],
+        [BASH, str(script)],
         capture_output=True,
         text=True,
         env=env,
@@ -182,7 +209,7 @@ exit 0
     }
 
     result = subprocess.run(
-        ["bash", str(script)],
+        [BASH, str(script)],
         capture_output=True,
         text=True,
         env=env,
@@ -258,7 +285,7 @@ echo '{"info":{"version":"0.4.13"}}'
     }
 
     result = subprocess.run(
-        ["bash", str(script)],
+        [BASH, str(script)],
         capture_output=True,
         text=True,
         env=env,
@@ -317,7 +344,7 @@ exit 0
     }
 
     result = subprocess.run(
-        ["bash", str(script)],
+        [BASH, str(script)],
         capture_output=True,
         text=True,
         env=env,
@@ -387,7 +414,7 @@ exit 0
     }
 
     result = subprocess.run(
-        ["bash", str(script)],
+        [BASH, str(script)],
         capture_output=True,
         text=True,
         env=env,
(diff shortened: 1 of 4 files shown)


zc277584121 (changes requested):
Thanks for addressing the native Windows test failure. The underlying issue is valid, but this head is not ready to merge:

1. Please rebase onto current `main`. This head is 32 commits behind, GitHub reports it as conflicting, and the hook tests have expanded substantially since the PR base.
2. Please keep Bash discovery in one shared test-infrastructure helper or fixture. Four identical module-local implementations will drift, and the current pytest configuration does not require duplicating this logic.
3. Please migrate every Bash subprocess consumer on current `main`. There are currently 38 literal-Bash call sites across five test modules (35 are in Windows-eligible tests); this head updates 13 old call sites across four modules and misses `tests/test_user_prompt_submit_hooks.py`.
4. Please make the failure mode actionable. Returning the literal `"bash"` when neither a compatible Bash nor a usable Git for Windows installation is found recreates the original `FileNotFoundError`. The resolver should also avoid selecting a WSL launcher or another incompatible shell merely because it is named `bash`.
5. Please add deterministic coverage for PATH/PATHEXT, the common `Git\\cmd\\git.exe` and `Git\\bin\\git.exe` layouts, paths containing spaces, shim/unsupported layouts, and no compatible Bash. Passing the resolved executable as one argv element is the right shape; the tests should lock that behavior down.
6. Please attach reproducible exact-head results for pytest launched from both PowerShell and cmd, pytest run inside Git Bash, a repository/install path containing spaces, and the no-Bash case, plus Linux and macOS regression results. Include the commands, Python/Git versions, relevant sanitized PATH/PATHEXT details, and complete pass/fail/skip counts.

Several invoked scripts also depend on tools such as `python3`, `pgrep`, `nohup`, or other POSIX utilities. Please state which current-main tests are expected to pass in native Windows environments and keep intentionally POSIX-only cases explicit. This does not require a broader runtime rewrite in this PR, but resolving Bash alone should not be presented as end-to-end Windows coverage when a test deterministically stops at the next missing tool.

Once the branch is rebased, the resolver is shared, all current consumers are covered, and the platform evidence is available, this should be straightforward to review again.

PR #669, opened 2026-08-06
fix: tolerate prose around maintenance JSON responses

### LINKED ISSUE #668: help: background maintenance tasks fail on non-OpenAI-native chat providers (minimax M3 returns thinking-prefixed text, not JSON)
---
title: "help: background maintenance tasks fail on non-OpenAI-native chat providers (minimax M3 returns thinking-prefixed text, not JSON)"
labels: "help wanted, documentation"
---

## Context

I'm running memsearch plugin for Claude Code with a Chinese LLM provider called `<provider-A>` (let's call it this for neutrality). The provider exposes an OpenAI-compatible endpoint at `https://<provider>.com/v1/chat/completions`, supports function calls, and returns standard OpenAI-format responses.

The **summarize** and **index** paths work fine with this provider. The plugin config:

```toml
[llm.providers.[provider-A]]
type = "openai"
model = "<provider-A>-chat"
base_url = "https://<provider>.com/v1"
api_key = "env:PROVIDER_API_KEY"

[plugins.claude-code.summarize]
provider = "[provider-A]"
model = "<provider-A>-chat"
```

→ Stop hook summaries land in `.memsearch/memory/YYYY-MM-DD.md` as expected.

## Problem

Background maintenance tasks (`project_review`, `user_profile`, `memory_to_skill`) **fail at runtime** with:

```
RuntimeError: Maintenance LLM did not return valid JSON: Expecting value: line 1 column 1 (char 0)
```

The state file `~/.memsearch/.maintenance-state.json` records:

```json
{
  "claude-code.user_profile": {
    "last_error": "RuntimeError: Maintenance LLM did not return valid JSON: Expecting value: line 1 column 1 (char 0)",
    "last_failed_at": "2026-08-05T07:36:10.405714Z"
  }
}
```

## Root Cause Analysis

Direct `curl` to the provider's chat completions endpoint (mirror of what memsearch's `maintenance.py` calls):

```bash
curl -X POST "https://<provider>.com/v1/chat/completions" \
  -H "Authorization: Bearer $TOKEN" \
  -H "Content-Type: application/json" \
  -d '{
    "model": "<provider-A>-chat",
    "messages": [{"role": "user", "content": "Return only JSON: {\"action\":\"none\"}"}],
    "max_tokens": 1000,
    "temperature": 0
  }'
```

The response:

```json
{
  "choices": [{
    "message": {
      "content": "{\"action\":\"none\"}",
      "role": "assistant"
    }
  }],
  "usage": {
    "completion_tokens_details": {"reasoning_tokens": 0}
  }
}
```

But with smaller `max_tokens` (default 200), the response gets **truncated mid-JSON**:

```json
{
  "message": {
    "content": "{\"action\":\"none\"}\""   ← trailing " breaks json.loads
  }
}
```

And with the **default** memsearch prompt (which builds a long `## Current Direction / Active Threads / Recent Progress ...` template), the model adds a `## Reasoning` block in front of the JSON:

```
## Reasoning
The user is asking me to update PROJECT.md based on recent journals...

## Result
{"action":"none"}
```

In either case, `json.loads(content)` in `maintenance.py:578` fails because the first non-whitespace character is `#` (from `## Reasoning`) or `"` (from truncated JSON).

## What I Tried (all failed)

| Attempt | Result |
|---|---|
| Set `max_tokens=1000` | Some prompts still hit ceiling (long journals + big prompt) |
| Custom prompt via `[prompts].project_review` and `[prompts].user_profile` forcing JSON output | Model still adds Chinese-language leading prose |
| System prompt `"Never output think tags, reasoning, or analysis. Output ONLY what the user requests."` | Worked for trivial prompts (10 chars), but ignored for real 18k-char prompts |
| Patch `maintenance.py` to strip ` ```json ... ``` ` fences | Already supported; my provider doesn't wrap in fences |
| Patch `maintenance.py` to strip `## Reasoning` blocks | Worked for some prompts, but the user-managed codebase (`/home/jade/.local/share/uv/tools/memsearch/...`) gets overwritten on every `uv tool install --force` |

## What I'm Asking For

I'd like **some combination** of:

1. **Pre-validation / tolerant parsing** in `maintenance.py`:
   - Strip leading `## Reasoning` / non-JSON prose before `json.loads`
   - On parse failure, retry the LLM call with a stricter prompt (e.g., "failed: only return JSON, no markdown")
   - Or fall back to `action: "none"` with the raw text logged to `.maintenance-state.json`

2. **Configurable request params** for the maintenance calls:
   - `max_tokens` (currently hardcoded or per-provider default)
   - `response_format: {type: "json_object"}` (OpenAI native JSON mode) — opt-in, since not all providers support it
   - `extra_body` passthrough for provider-specific knobs (e.g. `chat_template_kwargs` for Groq, `thinking` for some others)

3. **Documentation / examples**:
   - A worked example using a non-OpenAI-native provider (DeepSeek, Zhipu, MiniMax, etc.) with the exact recommended config in `~/.memsearch/config.toml`
   - A note that `summarize` is more permissive (plain text) but `maintenance` is strict JSON, so the failure modes differ

4. **Provider-aware defaults**:
   - Skip `maintenance` automatically when the provider name doesn't match a known JSON-strict list (or always run it in `strict_json` mode with a fallback)

My preferred fix would be (1) lenient parsing + (2) `max_tokens` + `response_format` config — the smallest scope that unblocks non-OpenAI-native providers.

## Environment

- memsearch 0.4.17
- Plugin installed via marketplace
- Provider: `<provider-A>` (OpenAI-compatible REST API)
- OS: Linux

## Workaround (current)

I disabled all `project_review` / `user_profile` / `memory_to_skill` maintenance tasks via `memsearch config set plugins.claude-code.<task>.enabled false`. This means I lose the auto-generated `PROJECT.md` / `USER.md` / `skill-candidates/`. I write `PROJECT.md` by hand when needed.

Happy to send a PR if you can point me to a maintainer preference for the approach (lenient parsing vs. config knob vs. docs).

Roovelrz:
## Summary

Fixes #668.

Maintenance tasks currently expect the complete LLM response to be valid JSON, apart from the existing fenced JSON handling. Some OpenAI-compatible providers return an otherwise valid maintenance result with leading reasoning or trailing text, causing `_parse_task_response` to reject the response.

This change makes maintenance response parsing tolerant of surrounding prose while preserving the existing validation and failure behavior.

## Changes

* Preserve support for plain JSON and fenced JSON responses
* Use `json.JSONDecoder().raw_decode()` to locate complete JSON objects within mixed text
* Prefer the final complete object containing a maintenance `action`
* Continue validating that `action` is either `none` or `replace`
* Ignore unrelated or invalid brace sequences before the actual result
* Preserve nested braces inside JSON string values
* Continue rejecting genuinely malformed or truncated JSON
* Avoid provider-specific parsing or silent fallback to `action: none`

## Tests

Added regression coverage for:

* [x] Existing plain JSON responses
* [x] Existing fenced JSON responses
* [x] Reasoning text before a valid maintenance result
* [x] Trailing text or an extra quote after a complete JSON object
* [x] Invalid brace sequences before the actual result
* [x] Multiple JSON objects where the final object is the maintenance result
* [x] Braces inside replacement content
* [x] Truncated JSON remaining an explicit error
* [x] Unsupported maintenance actions remaining an explicit error

## Validation

Focused parser tests:

```text
uv run python -m pytest tests/test_maintenance.py -k "parse_task_response" -v
```

Maintenance tests excluding the existing Windows path-handling failure:

```text
uv run python -m pytest tests/test_maintenance.py -k "not run_memory_command_allows_transcript_outside_roots" -v
```

Code quality checks:

```text
uv run ruff check src/memsearch/maintenance.py tests/test_maintenance.py
uv run ruff format --check src/memsearch/maintenance.py tests/test_maintenance.py
```

The full test suite was also executed on Windows:

```text
264 passed, 7 skipped, 29 failed, 8 errors
```

The remaining failures and errors are outside the changed response-parsing code paths and fall into Windows-specific or environment-sensitive areas, including:

* Bash-based Claude Code and Codex hook execution
* Windows symlink privilege requirements
* Windows path normalization and home-directory behavior
* The existing maintenance command path test
* Milvus Lite, which does not provide Windows wheels

The focused response parsing tests and the remaining maintenance tests pass locally.

## Scope

This PR intentionally remains limited to provider-agnostic response parsing.

It does not add:

* Configurable `max_tokens`
* `response_format`
* Provider-specific `extra_body` parameters
* Automatic retries
* Silent fallback to `action: none`

Those changes can be considered separately without expanding this focused parsing fix.

### DIFF
--- a/src/memsearch/maintenance.py
+++ b/src/memsearch/maintenance.py
@@ -583,21 +583,21 @@ def _parse_task_response(raw: str) -> dict[str, str]:
         cursor = 0
         while True:
             brace = text.find("{", cursor)
-            if brace < 0:
+            bracket = text.find("[", cursor)
+            starts = [position for position in (brace, bracket) if position >= 0]
+            if not starts:
                 break
+            start = min(starts)
 
             try:
-                candidate, end = decoder.raw_decode(
-                    text,
-                    brace,
-                )
+                candidate, end = decoder.raw_decode(text, start)
             except json.JSONDecodeError:
-                depth = 0
+                stack: list[str] = []
                 in_string = False
                 escaped = False
                 balanced_end: int | None = None
 
-                for index in range(brace, len(text)):
+                for index in range(start, len(text)):
                     char = text[index]
 
                     if in_string:
@@ -611,11 +611,14 @@ def _parse_task_response(raw: str) -> dict[str, str]:
 
                     if char == '"':
                         in_string = True
-                    elif char == "{":
-                        depth += 1
-                    elif char == "}":
-                        depth -= 1
-                        if depth == 0:
+                    elif char in "{[":
+                        stack.append(char)
+                    elif char in "}]":
+                        expected = "{" if char == "}" else "["
+                        if not stack or stack[-1] != expected:
+                            break
+                        stack.pop()
+                        if not stack:
                             balanced_end = index + 1
                             break
 
--- a/src/memsearch/maintenance.py
+++ b/src/memsearch/maintenance.py
@@ -572,10 +572,34 @@ def _parse_task_response(raw: str) -> dict[str, str]:
     match = re.search(r"```(?:json)?\s*(\{.*?\})\s*```", text, re.DOTALL)
     if match:
         text = match.group(1)
+
     try:
         data = json.loads(text)
-    except json.JSONDecodeError as e:
-        raise RuntimeError(f"Maintenance LLM did not return valid JSON: {e}") from e
+    except json.JSONDecodeError as initial_error:
+        decoder = json.JSONDecoder()
+        last_object: dict[str, Any] | None = None
+        last_action_object: dict[str, Any] | None = None
+
+        for brace in re.finditer(r"\{", text):
+            try:
+                candidate, _ = decoder.raw_decode(
+                    text,
+                    brace.start(),
+                )
+            except json.JSONDecodeError:
+                continue
+
+            if not isinstance(candidate, dict):
+                continue
+
+            last_object = candidate
+            if "action" in candidate:
+                last_action_object = candidate
+
+        data = last_action_object or last_object
+        if data is None:
+            raise RuntimeError(f"Maintenance LLM did not return valid JSON: {initial_error}") from initial_error
+
     action = data.get("action")
     if action not in {"none", "replace"}:
         raise RuntimeError("Maintenance LLM action must be 'none' or 'replace'")
--- a/src/memsearch/maintenance.py
+++ b/src/memsearch/maintenance.py
@@ -580,14 +580,52 @@ def _parse_task_response(raw: str) -> dict[str, str]:
         last_object: dict[str, Any] | None = None
         last_action_object: dict[str, Any] | None = None
 
-        for brace in re.finditer(r"\{", text):
+        cursor = 0
+        while True:
+            brace = text.find("{", cursor)
+            if brace < 0:
+                break
+
             try:
-                candidate, _ = decoder.raw_decode(
+                candidate, end = decoder.raw_decode(
                     text,
-                    brace.start(),
+                    brace,
                 )
             except json.JSONDecodeError:
+                depth = 0
+                in_string = False
+                escaped = False
+                balanced_end: int | None = None
+
+                for index in range(brace, len(text)):
+                    char = text[index]
+
+                    if in_string:
+                        if escaped:
+                            escaped = False
+                        elif char == "\\":
+                            escaped = True
+                        elif char == '"':
+                            in_string = False
+                        continue
+
+                    if char == '"':
+                        in_string = True
+                    elif char == "{":
+                        depth += 1
+                    elif char == "}":
+                        depth -= 1
+                        if depth == 0:
+                            balanced_end = index + 1
+                            break
+
+                if balanced_end is None:
+                    break
+
+                cursor = balanced_end
                 continue
+
+            cursor = end
 
             if not isinstance(candidate, dict):
                 continue
--- a/tests/test_maintenance.py
+++ b/tests/test_maintenance.py
@@ -230,6 +230,34 @@ The expected no-op shape is {"action":"none","reason":"example"}.
     }
 
 
+def test_parse_task_response_preserves_outer_action_with_nested_object() -> None:
+    raw = (
+        json.dumps(
+            {
+                "action": "replace",
+                "reason": "durable update",
+                "content": "# Project Memory",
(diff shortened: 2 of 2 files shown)


src/memsearch/maintenance.py
```diff
@@ -572,10 +572,34 @@ def _parse_task_response(raw: str) -> dict[str, str]:
...
     try:
         data = json.loads(text)
-    except json.JSONDecodeError as e:
-        raise RuntimeError(f"Maintenance LLM did not return valid JSON: {e}") from e
+    except json.JSONDecodeError as initial_error:
+        decoder = json.JSONDecoder()
+        last_object: dict[str, Any] | None = None
+        last_action_object: dict[str, Any] | None = None
+
+        for brace in re.finditer(r"\{", text):
```
yhmo: src/memsearch/maintenance.py line:583
Medium ---- This loop also starts `raw_decode` at braces nested inside an object that was already decoded. For example, with a top-level `replace` result containing `"meta":{"action":"none"}` plus trailing prose, the outer object is selected first and then the nested object overwrites it, so `run_due_tasks` records a successful no-op and will skip retrying the unchanged input. A truncated outer object can likewise be accepted if it contains a complete nested action object, contrary to the explicit truncation failure contract. Please use the returned end offset to skip braces inside each decoded candidate, or otherwise restrict candidates to standalone top-level objects.

Roovelrz: Thanks for catching this. I agree that scanning every opening brace can incorrectly treat a nested object as a later standalone maintenance result, and it can also weaken the truncated-response failure contract.

I’ll tighten the candidate selection so decoded top-level objects cannot be overwritten by nested objects, and add regression coverage for both cases:

* a valid outer `replace` result containing a nested object with `action: none`
* a truncated outer object containing a complete nested action object, which must still fail

I’ll keep the change limited to the parser behavior in this PR.

Roovelrz: Addressed in the latest commit.

I tightened the parser so a successfully decoded object advances the scan to its returned end offset, preventing nested objects from being reconsidered as standalone candidates. For malformed candidates, the scan now skips complete balanced brace blocks and stops on an unterminated outer object, so a complete nested action object cannot make a truncated response succeed.

I also added regression tests for both cases you identified:

* an outer `replace` result containing a nested `action: none` object remains `replace`
* a truncated outer object containing a complete nested action object still raises the explicit invalid-JSON error

The focused parser tests and the maintenance test set pass, along with Ruff checks.

PR #683, opened 2026-08-18
feat: add OrcaRouter embedding provider

XiaoHuo888-hue:
## Add OrcaRouter embedding provider

This registers [OrcaRouter](https://www.orcarouter.ai) as a named embedding provider in the provider registry, mirroring the existing `openai` provider. Instead of pointing the generic OpenAI provider at a custom `base_url`, users can set `embedding.provider = "orcarouter"` directly.

[OrcaRouter](https://www.orcarouter.ai) is an OpenAI-compatible model routing gateway that brings 150+ models from OpenAI, Anthropic, Google, DeepSeek, Qwen, MiniMax and xAI behind a single endpoint and API key. Beyond routing, it runs gateway-level, zero-trust security for AI agents on the same endpoint — screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes. This PR registers it as a named provider so users can opt in directly.

I'm an engineer on the OrcaRouter team.

### Files changed
- `src/memsearch/embeddings/orcarouter.py` (new) — `OrcaRouterEmbedding`, OpenAI-compatible, default base URL `https://api.orcarouter.ai/v1`, default model `openai/text-embedding-3-small`, key read from `ORCAROUTER_API_KEY` (override base URL via `ORCAROUTER_BASE_URL`).
- `src/memsearch/embeddings/__init__.py` — registered `orcarouter` in `_PROVIDERS`, `DEFAULT_MODELS`, `_INSTALL_HINTS`, and in the `base_url`/`api_key` pass-through of `get_provider()`.
- `src/memsearch/cli.py` — `config init` now lists `orcarouter` with its default model.
- `tests/test_embeddings_orcarouter.py` (new) — mirrors `test_embeddings_openai.py` (skipped without `ORCAROUTER_API_KEY`).
- Docs: provider table in `docs/home/configuration.md`, config-init prompt output in `docs/cli.md` and `docs/getting-started.md`, README config example, and the provider list in `CLAUDE.md`.

### Notes
- OrcaRouter model IDs are namespaced (e.g. `openai/text-embedding-3-small`); bare IDs like `text-embedding-3-small` return `503 model_not_found`, so the provider defaults to the namespaced form.
- The `openai` SDK is already a base dependency, so no new extras are required.

### Verification
- `ruff check` and `ruff format --check` pass on all touched files.
- `tests/test_embeddings_orcarouter.py` passes (5/5) against the live endpoint.
- The config/CLI/embedding test subset shows no regressions; the only failures are pre-existing Windows/Milvus-Lite environment issues present on `main` as well.

### DIFF
--- a/docs/home/configuration.md
+++ b/docs/home/configuration.md
@@ -74,6 +74,7 @@ memsearch index docs/ --exclude 'generated/**' --exclude '*.draft.md'
 |----------|---------|---------|-------|
 | **onnx** (default) | `pip install memsearch[onnx]` | No | Local, free, ~100MB model download |
 | openai | `pip install memsearch[openai]` | `OPENAI_API_KEY` | Best quality |
+| orcarouter | `pip install memsearch` | `ORCAROUTER_API_KEY` | OpenAI-compatible routing gateway |
 | google | `pip install memsearch[google]` | `GOOGLE_API_KEY` | Gemini embeddings |
 | voyage | `pip install memsearch[voyage]` | `VOYAGE_API_KEY` | High quality |
 | jina | `pip install memsearch[jina]` | `JINA_API_KEY` | jina-embeddings-v4, multilingual, long context |
--- a/src/memsearch/embeddings/__init__.py
+++ b/src/memsearch/embeddings/__init__.py
@@ -24,6 +24,7 @@ class EmbeddingProvider(Protocol):
 # Provider registry: name -> (module_path, class_name)
 _PROVIDERS: dict[str, tuple[str, str]] = {
     "openai": ("memsearch.embeddings.openai", "OpenAIEmbedding"),
+    "orcarouter": ("memsearch.embeddings.orcarouter", "OrcaRouterEmbedding"),
     "google": ("memsearch.embeddings.google", "GoogleEmbedding"),
     "voyage": ("memsearch.embeddings.voyage", "VoyageEmbedding"),
     "jina": ("memsearch.embeddings.jina", "JinaEmbedding"),
@@ -37,6 +38,7 @@ _PROVIDERS: dict[str, tuple[str, str]] = {
 # Kept here so callers can resolve the effective model without importing heavy deps.
 DEFAULT_MODELS: dict[str, str] = {
     "openai": "text-embedding-3-small",
+    "orcarouter": "openai/text-embedding-3-small",
     "google": "gemini-embedding-001",
     "voyage": "voyage-3-lite",
     "jina": "jina-embeddings-v4",
@@ -48,6 +50,7 @@ DEFAULT_MODELS: dict[str, str] = {
 
 _INSTALL_HINTS: dict[str, str] = {
     "openai": "pip install memsearch  (or: uv add memsearch)",
+    "orcarouter": "pip install memsearch  (or: uv add memsearch)",
     "google": 'pip install "memsearch[google]"  (or: uv add "memsearch[google]")',
     "voyage": 'pip install "memsearch[voyage]"  (or: uv add "memsearch[voyage]")',
     "jina": 'pip install "memsearch[jina]"  (or: uv add "memsearch[jina]")',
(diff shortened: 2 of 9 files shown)


src/memsearch/embeddings/__init__.py
```diff
@@ -24,6 +24,7 @@ async def embed(self, texts: list[str]) -> list[list[float]]: ...
 # Provider registry: name -> (module_path, class_name)
 _PROVIDERS: dict[str, tuple[str, str]] = {
     "openai": ("memsearch.embeddings.openai", "OpenAIEmbedding"),
+    "orcarouter": ("memsearch.embeddings.orcarouter", "OrcaRouterEmbedding"),
```
yhmo, line 27: src/memsearch/embeddings/__init__.py line:27
Medium ---- Registering `orcarouter` here makes it a valid `embedding.provider` value, but the plugin session-start hooks that map a provider to its required API key were not updated: `_required_env_var` in plugins/claude-code/hooks/session-start.sh:58-66 and plugins/codex/hooks/session-start.sh:41-49 only maps openai/google/voyage/jina/mistral to their env vars and returns "" for everything else. With `provider=orcarouter` and `ORCAROUTER_API_KEY` unset, the hook reports a healthy status line instead of the "ERROR: ... not set - memory search disabled" warning it shows for the other key-based providers, so plugin users get a silent broken config. The previous provider addition (PR #346) updated these hooks; please add `orcarouter) echo "ORCAROUTER_API_KEY" ;;` to both here.

docs/home/configuration.md
```diff
@@ -74,6 +74,7 @@ memsearch index docs/ --exclude 'generated/**' --exclude '*.draft.md'
 |----------|---------|---------|-------|
 | **onnx** (default) | `pip install memsearch[onnx]` | No | Local, free, ~100MB model download |
 | openai | `pip install memsearch[openai]` | `OPENAI_API_KEY` | Best quality |
+| orcarouter | `pip install memsearch` | `ORCAROUTER_API_KEY` | OpenAI-compatible routing gateway |
```
yhmo, line 77: docs/home/configuration.md line:77
Medium ---- This row keeps the provider table current, but two other surfaces that enumerate every embedding provider were not updated: docs/home/comparison.md:37 still claims "(8: openai / google / voyage / jina / mistral / ollama / local / onnx)" (now 9 providers; the count and list are stale after this PR), and docs/python-api.md:45 omits `orcarouter` from the `embedding_provider` parameter description. PR #346 updated both of these files when jina/mistral were added; please extend them here as well so the docs stay consistent for users discovering the new provider.

src/memsearch/embeddings/__init__.py
```diff
@@ -71,16 +74,18 @@ def get_provider(
...
     batch_size:
         Maximum number of texts per embedding API call.
         ``0`` means use the provider's built-in default.
     base_url:
-        Override the API base URL (currently only used by the openai provider).
+        Override the API base URL (currently only used by the openai and
+        orcarouter providers).
     api_key:
-        Override the API key (currently only used by the openai provider).
+        Override the API key (currently only used by the openai and
```
yhmo, line 87: src/memsearch/embeddings/__init__.py line:87
Low ---- This edited docstring claims the `api_key` override is "currently only used by the openai and orcarouter providers", but the same function also forwards `api_key` to the `jina` and `mistral` providers (`elif name in ("jina", "mistral") and api_key:` right below). Since these docstring lines are being touched anyway, consider correcting the claim (e.g. "used by the openai, orcarouter, jina and mistral providers") so the public docstring is accurate.

src/memsearch/embeddings/orcarouter.py
```diff
@@ -0,0 +1,93 @@
...
+"""
+
+from __future__ import annotations
+
+import os
+
+_DEFAULT_BASE_URL = "https://api.orcarouter.ai/v1"
+
+# Known output dimensions for OrcaRouter-routable embedding models.
+_KNOWN_DIMENSIONS: dict[str, int] = {
```
yhmo, line 21: src/memsearch/embeddings/orcarouter.py line:21
Low ---- This dimension table is a namespaced duplicate of the openai provider's `_KNOWN_DIMENSIONS` for the same underlying models, and it is already missing an entry the openai table has: `openai/text-embedding-ada-002` (1536). Any model not in this 2-row table makes `_detect_dimension` fall through to a synchronous, blocking trial embed inside `__init__`, so every `MemSearch()` construction (index/search/watch) for such a model pays a live network round-trip just to resolve the dimension - and since OrcaRouter advertises 150+ models, this fallback is the common case rather than the exception. Consider completing the table (add `openai/text-embedding-ada-002` and any other routable embedding models) or documenting the construction-time network dependency for custom models.

tests/test_embeddings_orcarouter.py
```diff
@@ -0,0 +1,56 @@
...
+
+These tests require ORCAROUTER_API_KEY to be set.
+They are skipped if the key is not available.
+"""
+
+import os
+
+import pytest
+
+pytestmark = pytest.mark.skipif(
```
yhmo, line 11: tests/test_embeddings_orcarouter.py line:11
Low ---- All five new tests are skipped whenever `ORCAROUTER_API_KEY` is unset, so in CI the new module executes zero tests. Two new code paths can be covered without any key or network: the missing-key `RuntimeError` raised in `OrcaRouterEmbedding.__init__` (orcarouter-specific - openai relies on the SDK to raise instead) and the no-network dimension lookup for models already in `_KNOWN_DIMENSIONS`. Adding those offline unit tests would give the new provider guaranteed coverage independent of a live endpoint.

Issue #653, opened 2026-07-27
Maintenance tasks read journals oldest-first, so truncation discards the recent ones


wessozi:
## Summary

`project_review` and `user_profile` build their prompt from the **oldest** journals and truncate the **newest** ones away. Once a memory directory exceeds `MAX_PROMPT_CHARS`, the maintenance tasks stop seeing recent sessions entirely — and because they still return a well-reasoned `action: "none"`, this fails silently and looks like "the model decided there was nothing to update".

Version: memsearch `0.4.16`, claude-code plugin `0.4.16`, `provider = "native"`.

## Root cause

Two behaviours combine, both in `memsearch/maintenance.py`:

1. **`_read_recent_journals()` (line 264)** sorts by mtime **ascending**, so the returned string starts with the *oldest* file:

   ```python
   files = sorted((p for p in input_dir.rglob("*.md") if p.is_file()),
                  key=lambda p: p.stat().st_mtime)[-max_files:]
   ```

   The `[-max_files:]` slice correctly selects the 12 most recent *files*, but the surviving list stays in oldest-first order.

2. **`_build_prompt()` (line 250)** then truncates by keeping the **head**:

   ```python
   if len(prompt) > MAX_PROMPT_CHARS:      # 80_000
       prompt = prompt[:MAX_PROMPT_CHARS] + "\n\n[truncated]\n"
   ```

Because the journals section is emitted oldest-first, head-truncation discards exactly the recent entries the task exists to summarise. The function is named `_read_recent_journals`; what reaches the model is the oldest.

## Reproduction

Any memory directory whose journals exceed ~80k characters. On my install:

```
2026-07-21.md    267,802 chars   <-- 3.3x the entire budget, on its own
2026-07-22.md      4,926
2026-07-23.md     56,023
2026-07-24.md    835,265
2026-07-25.md      7,560
2026-07-26.md    198,013
2026-07-27.md      8,573         <-- never reaches the model
                ----------
total          1,378,162 chars
```

The model sees roughly the first 30% of the oldest file and nothing after it. Running the task manually confirms it:

```
python3 <plugin>/scripts/maintenance-runner.py \
  --platform claude-code --project-dir <dir> --memsearch-dir <dir>/.memsearch \
  --force --json-output
```

```json
[{"task": "project_review", "action": "none",
  "reason": "Recent journal entries reiterate decisions and progress already
             captured in the existing output file ...",
  "skipped": false}]
```

[...]

PR #691, opened 2026-08-19
fix: preserve latest session memory within context budget

### LINKED ISSUE #684: Claude Code SessionStart: tail-selected recent-memory preview is head-truncated by the harness, discarding the newest entries (same class as #653)
Thanks for memsearch — the Claude Code plugin's recent-memory injection is genuinely useful day to day. Filing this because I hit a subtle interaction that quietly defeats the feature's purpose on active projects.

Verified against `main` @ 36216d6 (also present in v0.4.17).

## Summary

`plugins/claude-code/hooks/session-start.sh` builds its "Recent Memory" context by taking the last `max_lines` lines of the 2 most recent daily journal files via `tail`. On a sufficiently active project the assembled block exceeds Claude Code's threshold for inline hook context, so the harness truncates it to a ~2KB **head** preview (writing the full payload to a file and injecting a pointer).

The hook selects the *newest* lines, but the harness then keeps only the *oldest* portion of that selection. So the entries most likely to matter for a session that is just starting are exactly the ones dropped. The output still renders as a well-formed "Recent Memory" block, so nothing signals that content was cut.

This is the same failure mode as #653 (maintenance tasks reading journals oldest-first), which was accepted and fixed for that code path. Notably, the discussion there concluded that reversing *file* order fixes truncation across files but "does not preserve the latest entries within a file" — that within-file case is what this issue is about, on a different code path.

## Where

`plugins/claude-code/hooks/session-start.sh`:

- Definition, lines 131-132:
  ```bash
  _recent_memory_preview() {
    local file="$1" max_lines="${2:-40}"
  ```
- `tail` selection, line 165:
  ```bash
  ' "$file" 2>/dev/null | tail -n "$max_lines" || true
  ```
- Call site with hardcoded window, line 229:
  ```bash
  content=$(_recent_memory_preview "$f" 40)
  ```

## Measurements

Against a real project's daily journal (avg 289 bullets/day, avg bullet length 186 chars), assembling the 2 most recent files at different `max_lines` values:

| max_lines | assembled size | outcome |
|---|---|---|
| 40 (current default) | 12,032 B | truncated; injected slice is the oldest part of the window |
| 20 | 6,948 B | truncated |
| 12 | 3,556 B | truncated |
| 8 | 2,402 B | truncated |
| 6 | 1,746 B | fits uncut; injected content is genuinely the newest |

Counterintuitively a *smaller* window is strictly better here: it is the only value tested that stays under the truncation threshold, and therefore the only one where the injected context is actually the most recent entries rather than the oldest.

## Suggestions

Happy to send a PR for any of these:

1. **Make the preview window configurable** rather than hardcoded at line 229 — e.g. `plugins.claude-code.session_start.max_lines`, consistent with existing keys like `plugins.claude-code.summarize.enabled`. `_recent_memory_preview` already takes the value as a parameter, so this is mostly plumbing a config read to the call site.
2. **Consider budgeting by bytes rather than lines.** Line length varies a lot with summary verbosity, so a line count cannot reliably stay under a byte-based truncation threshold the way a byte budget would.
3. **Consider assembling so the newest content survives truncation** — ordering or trimming such that a head-truncation keeps the most recent entries.

## Note on workarounds

A local edit to this file is not a viable stopgap: it lives in a plugin cache directory that is not version-controlled and gets overwritten by `claude plugin update`. A local fix silently reverts on the next update, taking the session-start context with it.

Happy to provide more detail or test a patch against the same project.

Roovelrz:
## Summary

Fixes #684.

Claude Code SessionStart currently builds the Recent Memory context from the two most recent daily journals and selects the latest entries from each journal. However, the assembled `additionalContext` can still exceed the Claude Code inline hook context limit.

When that happens, the downstream harness keeps a head preview of the payload. This means memsearch first selects recent entries, but the newest part of that selection can still be discarded before it reaches the session.

This is the same class of recency-preservation issue addressed for maintenance prompts in #653 and #656, but on the Claude Code SessionStart path.

This change makes the SessionStart recent-memory preview byte-budget aware before it is passed to the harness. When the available context is limited, newer journal content is prioritized and older content is dropped first.

## Changes

* Add an explicit byte budget for Claude Code SessionStart recent-memory context
* Keep the existing recent-session filtering and line-based candidate window
* Apply the byte budget before emitting `additionalContext`
* Preserve complete recent lines instead of relying on downstream head truncation
* Prioritize the newest daily journal when the combined recent-memory context exceeds the available budget
* Keep the final Recent Memory block within the configured internal size limit
* Preserve valid UTF-8 content when trimming multibyte text
* Leave the existing two-journal selection behavior unchanged

The fix is intentionally scoped to the Claude Code SessionStart path. It does not change maintenance prompt budgeting, memory indexing, or other plugin implementations.

## Why byte budgeting

The existing preview uses a fixed line count, but journal entries can vary significantly in length. A smaller line count reduces the likelihood of downstream truncation but cannot guarantee that the resulting payload stays within a byte-based context limit.

Budgeting the generated Recent Memory block by bytes allows memsearch to control truncation before handing the context to Claude Code and ensures that truncation follows memsearch's recency policy rather than the harness's head-preview behavior.

## Recency behavior

When the available budget is insufficient for all selected recent memory, the priority is:

1. Newest content from the newest daily journal
2. Earlier content from the newest daily journal
3. Content from the previous daily journal when space remains

This keeps the information most relevant to a newly started session while preserving the existing chronological order of the content that survives.

If the newest candidate entry itself exceeds the available budget, SessionStart does not fall back to an older journal, preserving recency priority rather than injecting stale context.

## Tests

Added regression coverage for:

* An oversized single daily journal where the latest entry must remain in `additionalContext`
* Multiple recent journals where the newest journal must take priority when the shared budget is exhausted
* Keeping the final Recent Memory context within the byte budget
* UTF-8 journal content remaining valid after byte-aware trimming
* Existing behavior that skips empty sessions and ignores non-daily journal files

Commands run:

```text
uv run python -m pytest tests/test_claude_hooks.py -v
uv run ruff check tests/test_claude_hooks.py
uv run ruff format --check tests/test_claude_hooks.py
uv run python -m pytest
```

## Test plan

* [x] Oversized recent memory keeps the latest journal entry
* [x] Older entries are removed before newer entries when the budget is exceeded
* [x] The newest daily journal is prioritized over the previous daily journal
* [x] The emitted Recent Memory block stays within the internal byte budget
* [x] UTF-8 content remains valid after trimming
* [x] Existing Claude Code SessionStart hook tests pass
* [x] Ruff lint passes
* [x] Ruff format check passes
* [x] Full test suite passes

### DIFF
--- a/plugins/claude-code/hooks/session-start.sh
+++ b/plugins/claude-code/hooks/session-start.sh
@@ -129,6 +129,8 @@ PROJECT_BASENAME=$(basename "${CLAUDE_PROJECT_DIR:-.}")
 COLLECTION_DESC="${PROJECT_BASENAME} | ${PROVIDER}/${MODEL:-default}"
 
 RECENT_MEMORY_MAX_LINES=40
+# Claude Code head-truncates inline hook context at roughly 2 KB.
+# Keep a safety margin below the observed threshold from #684.
 RECENT_MEMORY_MAX_BYTES=1800
 
 _byte_len() {
@@ -164,12 +166,13 @@ _tail_lines_within_bytes() {
 
 _recent_memory_preview() {
   local file="$1" max_lines="${2:-40}" max_bytes="${3:-0}"
+  local candidate trimmed
 
   if [ "$max_bytes" -le 0 ]; then
     return 0
   fi
 
-  awk '
+  candidate=$(awk '
     function flush_section() {
       if (section_len > 0 && has_body) {
         for (i = 1; i <= section_len; i++) {
@@ -201,7 +204,23 @@ _recent_memory_preview() {
     END {
       flush_section()
     }
-  ' "$file" 2>/dev/null | tail -n "$max_lines" | _tail_lines_within_bytes "$max_bytes" || true
+  ' "$file" 2>/dev/null | tail -n "$max_lines" || true)
+
+  # No useful recent-memory content in this journal. The caller may try
+  # an older journal.
+  if [ -z "$candidate" ]; then
+    return 0
+  fi
+
+  trimmed=$(printf '%s\n' "$candidate" | _tail_lines_within_bytes "$max_bytes")
+
+  # Candidate content exists, but even its newest complete line does not fit.
+  # Signal the caller not to fall back to an older journal.
+  if [ -z "$trimmed" ]; then
+    return 2
+  fi
+
+  printf '%s' "$trimmed"
 }
 
 # The session heading is written lazily by stop.sh on the first
@@ -260,6 +279,7 @@ recent_files=$(find "$MEMORY_DIR" -maxdepth 1 -type f -name "$DAILY_JOURNAL_PATT
 
 if [ -n "$recent_files" ]; then
   context="# Recent Memory\n\n"
+  has_recent_memory=false
   while IFS= read -r f; do
     [ -z "$f" ] && continue
     basename_f=$(basename "$f")
@@ -271,14 +291,26 @@ if [ -n "$recent_files" ]; then
     fi
     # Extract recent non-empty session sections. Legacy journals may contain
     # empty headings, but they do not carry useful context.
-    content=$(_recent_memory_preview "$f" "$RECENT_MEMORY_MAX_LINES" "$content_budget")
-    if [ -n "$content" ]; then
-      context+="$file_header$content\n\n"
+    content=""
+    if content=$(_recent_memory_preview "$f" "$RECENT_MEMORY_MAX_LINES" "$content_budget"); then
+      if [ -n "$content" ]; then
+        context+="$file_header$content\n\n"
+        has_recent_memory=true
+      fi
+    else
+      preview_status=$?
+      if [ "$preview_status" -eq 2 ]; then
+        break
+      fi
     fi
     if [ "$(_byte_len "$context")" -ge "$RECENT_MEMORY_MAX_BYTES" ]; then
       break
     fi
   done <<< "$recent_files"
+
+  if [ "$has_recent_memory" != true ]; then
+    context=""
+  fi
 fi
 
 # Note: Detailed memory search is handled by the memory-recall skill (pull-based).
--- a/plugins/claude-code/hooks/session-start.sh
+++ b/plugins/claude-code/hooks/session-start.sh
@@ -128,8 +128,47 @@ fi
 PROJECT_BASENAME=$(basename "${CLAUDE_PROJECT_DIR:-.}")
 COLLECTION_DESC="${PROJECT_BASENAME} | ${PROVIDER}/${MODEL:-default}"
 
+RECENT_MEMORY_MAX_LINES=40
+RECENT_MEMORY_MAX_BYTES=1800
+
+_byte_len() {
+  LC_ALL=C printf '%s' "$1" | wc -c | tr -d '[:space:]'
+}
+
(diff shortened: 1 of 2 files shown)


plugins/claude-code/hooks/session-start.sh
```diff
@@ -128,8 +128,47 @@ fi
...
+      sizes[NR] = length($0) + 1
+    }
+
+    END {
+      total = 0
+      start = NR + 1
+
+      for (i = NR; i >= 1; i--) {
+        if (total + sizes[i] > budget) {
+          break
```
yhmo, line 154: plugins/claude-code/hooks/session-start.sh line:152
Medium ---- When the newest daily journal's newest line alone exceeds the remaining content budget, this loop breaks before accepting it, so `_tail_lines_within_bytes` prints nothing and the whole newest journal is dropped; the loop then injects the *older* journal's content instead, inverting the PR's stated priority ("newest content from the newest daily journal" first). Verified by simulation: a single >~1750-byte bullet in the newest journal yields a payload containing only the older journal. The new tests never exercise this drop-entirely path. Consider keeping the newest section heading (or nothing at all from older journals) when the newest journal cannot fit, or documenting this fallback as intentional.

plugins/claude-code/hooks/session-start.sh
```diff
@@ -128,8 +128,47 @@ fi
 PROJECT_BASENAME=$(basename "${CLAUDE_PROJECT_DIR:-.}")
 COLLECTION_DESC="${PROJECT_BASENAME} | ${PROVIDER}/${MODEL:-default}"
 
+RECENT_MEMORY_MAX_LINES=40
+RECENT_MEMORY_MAX_BYTES=1800
```
yhmo, line 134: plugins/claude-code/hooks/session-start.sh line:132
Low ---- `RECENT_MEMORY_MAX_BYTES=1800` is a magic constant with no in-code rationale. Issue #684's measurements put the harness truncation threshold around 2KB (1,746 B fits, 2,402 B truncated), which is the entire justification for this value; a future maintainer has no basis in the code to adjust it. Suggest a one-line comment citing the ~2KB inline-hook context limit so the budget is discoverable and reviewable.

plugins/claude-code/hooks/session-start.sh
Roovelrz: Thanks for catching this. I updated the preview path to distinguish between a journal with no usable recent content and one where candidate content exists but its newest complete line cannot fit the remaining byte budget. In the latter case, SessionStart now stops instead of falling back to an older journal, preserving the stated recency priority. I also added a regression test covering a single oversized newest entry.

plugins/claude-code/hooks/session-start.sh
Roovelrz: Agreed. I added an in-code comment documenting that the 1800-byte budget keeps a safety margin below the roughly 2 KB inline-hook truncation threshold observed in #684.

plugins/claude-code/hooks/session-start.sh
yhmo: The issue is confirmed fixed.

plugins/claude-code/hooks/session-start.sh
yhmo: The issue is confirmed fixed.

Issue #600, opened 2026-06-29
Windows: kill_orphaned_index can't reap native index processes (Git Bash) -> unbounded accumulation / OOM


sdkkds:
## Summary

On Windows (hooks running under Git Bash), `kill_orphaned_index` in `hooks/common.sh` cannot reap the `memsearch index` processes spawned by the Stop hook. It reaps via Linux `pgrep`/`kill`, but the index it spawns is a **native** Windows `python.exe` (installed via Scoop/pip), which `pgrep` cannot see. The Stop hook spawns a fresh index every turn, so on Windows they are never reaped and accumulate until the machine runs out of memory.

## Relationship to existing issues (not a duplicate)

- **Not #533** (`Indexing leaks ~9 MB anon-rss per chunk in single process`). That is RSS growth *within a single* index process — "per-process restart releases memory normally." This bug is the inverse: whole `memsearch index` **processes** accumulate because `kill_orphaned_index` never reaps them on Windows, so the restart that would release memory never happens.
- Same under-tested Windows/Git-Bash hook path as #546 (Stop hook Unicode crash) and #75 (watcher on Windows), but a distinct failure (process reaping, not decode/watch).

## Environment

- Windows 11; Claude Code hooks executed under Git Bash (MSYS2)
- memsearch plugin **0.4.11**, server mode (Milvus at `localhost:19530`, WSL2 Docker)
- embedding: onnx `bge-m3` int8
- memsearch installed as a native Windows app (Scoop Python), so `command -v memsearch` in Git Bash resolves to `…\scoop\shims\memsearch.exe`

## Observed

After one long session:

- **10 live** `python.exe … memsearch.exe index <MEMORY_DIR> --collection ms_…` processes, **~4.5 GB** combined working set.
- Oldest was **24.7 hours** old — it survived a `SessionEnd` and a full Claude Code restart, proving nothing reaps it.
- The resulting memory pressure crashed the browser twice and Claude Code once.

The singleton `memsearch watch` process is fine (one process); the leak is purely the per-turn Stop-hook index.

## Root cause

- `hooks/stop.sh` runs on every turn and ends with:
  ```bash
  kill_orphaned_index
  run_memsearch index "$MEMORY_DIR"
  ```
[...]

Issue #658, opened 2026-07-30
Windows/Git Bash: stop_watch orphan sweep is a no-op (no pgrep) - duplicate watch processes accumulate


imbhat:
On Windows (Git for Windows bash), the watch-singleton logic in the Claude Code plugin's `hooks/common.sh` can leak duplicate `memsearch watch` processes on the same collection, because the orphan sweep in `stop_watch()` depends on `pgrep`, which Git for Windows does not ship.

## Observed

Windows 11, Git for Windows bash (Claude Code hook environment), plugin v0.4.16, Milvus standalone (http URI, so the persistent `watch` path is active).

Two live watchers on the same project/collection, started ~8 minutes apart:

```
Administ  1733  1  15:19:33 C:\...\memsearch.exe watch E:/IMBhat_AI_OS/.memsearch/memory --collection ms_imbhat_ai_os_b16e7dbd ...
Administ  2207  1  15:27:05 C:\...\memsearch.exe watch E:/IMBhat_AI_OS/.memsearch/memory --collection ms_imbhat_ai_os_b16e7dbd ...
```

`.watch.pid` contained only the newer pid (2207). On a previous day the pidfile also held a pid with no matching live process (stale), so the pidfile does get out of sync with reality in this environment.

## Why the singleton fails on Windows

`stop_watch()` has two lines of defense, and both have Windows problems:

1. **Pidfile kill** works when the pidfile is accurate — I verified from a separate bash instance that `kill <msys-pid>` does terminate a native `.exe` started via `nohup ... &`, and `kill -0` sees it. But it only protects the single pid recorded last; any watcher that escapes the pidfile (concurrent session start racing between `stop_watch`'s `rm -f` and `echo $! >`, a crashed session, a clobbered pidfile) is invisible to it.
2. **Orphan sweep** is the layer designed to catch escapees — `pgrep -f "memsearch watch $MEMORY_DIR"` — but Git for Windows has **no `pgrep`** (`command -v pgrep` → nothing), so `orphans=` is always empty and the `2>/dev/null || true` hides the failure. On Linux this sweep would have reaped the duplicate; on Windows duplicates accumulate until manually killed.

Also relevant for anyone debugging this: `echo $!` records an **MSYS pid**, not a Windows pid, so `.watch.pid` can't be checked with Task Manager / `Get-Process -Id` — a healthy watcher looks dead from the Windows side. (And `setsid` is also absent, so the `nohup` fallback path is the one running.)

## Suggested fix

[...]

PR #693, opened 2026-08-21
fix(hooks): keep the watch registry outside the watched tree

### LINKED ISSUE #692: Claude Code plugin: watch pidfile lives inside the watched dir, so removing a git worktree orphans its watcher permanently (34 leaked trees / 6.7 GB observed)
## Summary

Both of `stop_watch`'s recovery handles are keyed to the directory being watched. When that
directory is deleted — the normal end of a `git worktree remove` — the watcher becomes
unreachable by **both** mechanisms and runs forever. Each leak is a 3-process tree
(`bash` wrapper → `python3` → `tee`).

This is platform-independent and reproduces on macOS with `pgrep` present, so it is
distinct from #658 / #600 (Git Bash lacking `pgrep`). It is the same *function*, a different
cause: there the sweep can't run, here the sweep runs and is aimed at the wrong place.

Measured on one developer machine that uses one worktree per task:

| | |
|---|---|
| live watcher trees | 39 |
| whose watched directory no longer exists | **34** |
| total processes | 115 |
| resident memory | ~6.7 GB |
| age of oldest orphan | 9d 10h |

The host was at 95% swap utilisation. Reaping the 34 returned ~1.7 GB immediately and the OS
subsequently shrank the swapfile from 36 GB to 6 GB.

## Mechanism

In `plugins/claude-code/hooks/common.sh` @ HEAD:

```sh
MEMSEARCH_DIR="${MEMSEARCH_DIR:-$_PROJECT_DIR/.memsearch}"   # L41
MEMORY_DIR="$MEMSEARCH_DIR/memory"                           # L42
WATCH_PIDFILE="$MEMSEARCH_DIR/.watch.pid"                    # L292
```

```sh
stop_watch() {
  # 1. Kill the process recorded in pidfile
  if [ -f "$WATCH_PIDFILE" ]; then ... fi        # L305 — pidfile lives INSIDE the watched tree

  # 2. Sweep for orphaned watch processes targeting this MEMORY_DIR
  orphans=$(pgrep -f "memsearch watch $MEMORY_DIR" ...)   # L319 — scoped to THIS session's dir
}
```

Handle 1 — the pidfile is stored inside the very resource it manages, so it is deleted along
with it. This is a PID-file placement that only works in the case where cleanup isn't needed.

Handle 2 — the sweep is scoped by `$MEMORY_DIR`, i.e. by *the identity that was just
destroyed*. No future session will ever again have `MEMORY_DIR` equal to a removed worktree's
path, so nothing ever searches for that pattern again. (The orphan's own `argv` still contains
the deleted path, so a `pgrep` **would** match it — but no caller ever issues that query.)

With `start_watch` launching under `setsid`, the process survives its parent, and the two
handles that could reap it are gone and never re-aimed.

A second, smaller path: `stop_watch` only runs from `SessionEnd`. Any session killed without
that hook firing (lid close, SIGKILL, OOM) leaks one too — but that one is still recoverable,
because its directory usually still exists. The worktree case is the unrecoverable one.

## Reproduction

```sh
git worktree add ../wt-demo -b demo
cd ../wt-demo
# start a Claude Code session here, let the plugin spawn its watcher, exit cleanly
pgrep -fl "memsearch watch .*wt-demo"        # -> running, as expected

cd - && git worktree remove ../wt-demo
pgrep -fl "memsearch watch .*wt-demo"        # -> STILL RUNNING, now unreachable
```

Nothing in any subsequent session will reap it. Repeat per task branch to reproduce the
accumulation above.

## Why it's easy to miss

The processes appear in `ps`/Activity Monitor as bare `python3.11` with the script buried in
argv, so they don't read as attributable to the plugin. I initially could not trace them at
all — that's what started this investigation.

## Suggested fix

Move the pidfile out of the watched tree and make the sweep independent of the current
session's directory. Something like:

```sh
WATCH_STATE_DIR="${XDG_STATE_HOME:-$HOME/.local/state}/memsearch/watchers"
WATCH_PIDFILE="$WATCH_STATE_DIR/$(printf '%s' "$MEMSEARCH_DIR" | shasum | cut -c1-16).pid"
```

with the file recording both the PID **and** its target directory. Then a sweep at
`SessionStart` can iterate every pidfile, and reap any whose recorded target no longer
exists — which covers the worktree case *and* the lid-close case, without needing the dead
session to have run any cleanup itself. Keying on a hash rather than the path keeps the
filename stable and filesystem-safe.

A cheaper partial mitigation, if the state-dir move is unwanted: broaden handle 2 to
`pgrep -f "memsearch watch "` and reap any match whose target directory is absent. That
requires no new state, though it is less precise and would need care around concurrent
sessions.

## Workaround in the meantime

For anyone hitting this: a `SessionStart` hook that enumerates `memsearch watch` processes,
resolves each one's target directory from argv, and kills the trees whose target is gone.
Worth reaping the whole process group — each watcher is a 3-process tree, and killing only
the Python leaves the wrapper and `tee` behind.

Happy to open a PR for the state-dir approach if that direction is agreeable.

## Environment

- memsearch plugin 0.3.4 (bug also verified present at `main` HEAD, `common.sh` as quoted)
- macOS 15 / Darwin 25.5.0, Apple silicon
- Python 3.11.14, installed via `uv tool`
- Claude Code, one git worktree per task

samclams:
Fixes #692.

## The problem

`WATCH_PIDFILE` lives at `$MEMSEARCH_DIR/.watch.pid` — inside the very tree it manages. When that tree goes away (`git worktree remove`, deleting a scratch project), the only handle to the running watcher is destroyed along with it.

The fallback sweep in `stop_watch` cannot recover it either:

```sh
orphans=$(pgrep -f "memsearch watch $MEMORY_DIR" 2>/dev/null || true)
```

That query is scoped by the same `MEMORY_DIR` that was just deleted. No future session ever has that `MEMORY_DIR` again, so no future session ever issues a query that could match the orphan. The watcher is launched detached (`setsid`/`nohup`), so it survives indefinitely — the orphan's own `argv` still contains the path, but nothing ever looks for it.

Measured on one macOS laptop before the fix: **34 orphaned watcher trees, 115 processes, ~6.7 GB resident, the oldest 9 days old.** Reaping them returned ~1.7 GB and shrank swap from 36 GB to 6 GB.

## The fix

A handle used to clean something up has to outlive the thing being cleaned. This moves the handle out:

- Each watcher is recorded in `${XDG_STATE_HOME:-$HOME/.local/state}/memsearch/watchers/<pid>.pid`, containing the pid and its target directory. Overridable via `MEMSEARCH_STATE_DIR`.
- `start_watch` calls `reap_stale_watchers`, which kills any registered watcher whose target directory no longer exists — the removed-worktree case, reached from any project, not just the deleted one.
- `stop_watch` also clears registry entries for the current `MEMORY_DIR`.

Compatibility:

- The legacy in-tree pidfile is still written and still honoured, so upgrading does not strand a watcher an older version started.
- The existing `pgrep` sweep is unchanged.
- Entries are named by pid rather than a hash of the path, to avoid depending on `sha256sum`/`shasum`, which differ across platforms.
- PID reuse is guarded best-effort by checking the live process's `argv` before signalling. When the process cannot be inspected at all it is treated as ours — that is the current unconditional behaviour, not a new risk.

### Also fixes the Windows reports

`reap_stale_watchers` uses no `pgrep`. That makes it a path to recovery on Windows/Git Bash, where #658 and #600 report watchers accumulating precisely because `stop_watch`'s sweep needs `pgrep` and cannot find it. Different cause from #692, same symptom, and this reaper covers both.

## Both plugins

`plugins/claude-code/hooks/common.sh` and `plugins/codex/hooks/common.sh` carried the identical bug. Both are fixed, and the shared region is byte-identical between them.

## Tests

`tests/test_claude_hooks.py` gains three tests, parametrized over both plugins:

- The registry path resolves outside the project tree and outside `MEMSEARCH_DIR` — the invariant the bug violated.
- **Positive control:** target directory removed → the watcher is reaped and its entry cleared.
- **Negative control:** target directory present → the watcher survives and keeps its entry. This is the load-bearing one; without it, a reaper that killed unconditionally would pass the positive control.

Each control was verified by mutation: forcing the target-exists branch to `false` fails only the positive control, forcing it to `true` fails only the negative control.

Full suite green (`uv run python -m pytest`), `ruff check` and `ruff format --check` clean, `shellcheck` reports no new findings against the pristine files.

### DIFF
--- a/plugins/claude-code/hooks/common.sh
+++ b/plugins/claude-code/hooks/common.sh
@@ -289,7 +289,16 @@ kill_orphaned_index() {
 
 # --- Watch singleton management ---
 
+# Legacy pidfile. It lives inside the tree it manages, so it is destroyed along
+# with that tree (e.g. `git worktree remove`) and the watcher is left running
+# with no handle to reap it. Still read below so upgrading does not strand a
+# watcher that an older version started.
 WATCH_PIDFILE="$MEMSEARCH_DIR/.watch.pid"
+
+# Registry of running watchers, kept OUTSIDE any watched tree so that deleting a
+# project directory cannot take the only handle to its watcher with it. One file
+# per watcher, named for its PID, containing "<pid>\n<target dir>".
+WATCH_REGISTRY_DIR="${MEMSEARCH_STATE_DIR:-${XDG_STATE_HOME:-$HOME/.local/state}/memsearch}/watchers"
 
 # Kill a process and its entire process group to avoid orphans
 _kill_tree() {
@@ -298,23 +307,103 @@ _kill_tree() {
   kill -- -"$pid" 2>/dev/null || kill "$pid" 2>/dev/null || true
 }
 
+# Best-effort guard against PID reuse: a recorded PID may have been recycled by
+# an unrelated process before we get to signal it. Succeeds when the process
+# cannot be inspected at all, which is the unconditional behaviour this
+# replaces rather than a regression.
+_watch_pid_is_ours() {
+  local pid="$1" cmd=""
+  cmd="$(ps -o args= -p "$pid" 2>/dev/null || true)"
+  case "$cmd" in
+    "") return 0 ;;
+    *memsearch*) return 0 ;;
+    *) return 1 ;;
+  esac
+}
+
+_watch_registry_record() {
+  local pid="$1" target="$2"
+  mkdir -p "$WATCH_REGISTRY_DIR" 2>/dev/null || return 0
+  printf '%s\n%s\n' "$pid" "$target" > "$WATCH_REGISTRY_DIR/$pid.pid" 2>/dev/null || true
+}
+
+_watch_registry_forget() {
+  rm -f "$WATCH_REGISTRY_DIR/$1.pid" 2>/dev/null || true
+}
+
+# Reap watchers whose target directory no longer exists. This is the
+# removed-worktree case, which neither the in-tree pidfile (deleted with the
+# tree) nor a "$MEMORY_DIR"-scoped pgrep (no live session ever has that
+# MEMORY_DIR again) can reach. Uses no pgrep, so it also covers platforms where
+# pgrep is unavailable.
+reap_stale_watchers() {
+  [ -d "$WATCH_REGISTRY_DIR" ] || return 0
+  local entry pid target
+  for entry in "$WATCH_REGISTRY_DIR"/*.pid; do
+    [ -f "$entry" ] || continue
+    pid="$(sed -n '1p' "$entry" 2>/dev/null || true)"
+    target="$(sed -n '2p' "$entry" 2>/dev/null || true)"
+    if [ -z "$pid" ]; then
+      rm -f "$entry" 2>/dev/null || true
+      continue
+    fi
+    if ! kill -0 "$pid" 2>/dev/null; then
+      rm -f "$entry" 2>/dev/null || true
+      continue
+    fi
+    if [ -n "$target" ] && [ ! -d "$target" ]; then
+      if _watch_pid_is_ours "$pid"; then
+        _kill_tree "$pid"
+      fi
+      rm -f "$entry" 2>/dev/null || true
+    fi
+  done
+}
+
+# Kill any registered watcher whose target is this MEMORY_DIR.
+_stop_registered_watchers_for_memory_dir() {
+  [ -d "$WATCH_REGISTRY_DIR" ] || return 0
+  local entry pid target
+  for entry in "$WATCH_REGISTRY_DIR"/*.pid; do
+    [ -f "$entry" ] || continue
+    pid="$(sed -n '1p' "$entry" 2>/dev/null || true)"
+    target="$(sed -n '2p' "$entry" 2>/dev/null || true)"
+    if [ -z "$pid" ]; then
+      rm -f "$entry" 2>/dev/null || true
+      continue
+    fi
+    if [ "$target" = "$MEMORY_DIR" ]; then
+      if kill -0 "$pid" 2>/dev/null && _watch_pid_is_ours "$pid"; then
+        _kill_tree "$pid"
+      fi
+      rm -f "$entry" 2>/dev/null || true
+    fi
+  done
+}
+
 # Stop the watch process: pidfile first, then sweep for orphans
 stop_watch() {
   # Skip watch management in child claude -p processes (e.g. stop.sh summarization)
   if [ "${MEMSEARCH_NO_WATCH:-}" = "1" ]; then
     return 0
   fi
-  # 1. Kill the process recorded in pidfile
+  # 1. Kill the process recorded in the legacy in-tree pidfile
   if [ -f "$WATCH_PIDFILE" ]; then
     local pid
-    pid=$(cat "$WATCH_PIDFILE" 2>/dev/null)
+    pid="$(cat "$WATCH_PIDFILE" 2>/dev/null || true)"
     if [ -n "$pid" ] && kill -0 "$pid" 2>/dev/null; then
       _kill_tree "$pid"
+    fi
+    if [ -n "$pid" ]; then
+      _watch_registry_forget "$pid"
     fi
     rm -f "$WATCH_PIDFILE"
   fi
 
-  # 2. Sweep for orphaned watch processes targeting this MEMORY_DIR
+  # 2. Kill any registered watcher for this MEMORY_DIR
+  _stop_registered_watchers_for_memory_dir
+
+  # 3. Sweep for orphaned watch processes targeting this MEMORY_DIR
   local orphans
   orphans=$(pgrep -f "memsearch watch $MEMORY_DIR" 2>/dev/null || true)
   if [ -n "$orphans" ]; then
@@ -330,6 +419,10 @@ start_watch() {
   if [ "${MEMSEARCH_NO_WATCH:-}" = "1" ]; then
     return 0
   fi
+
+  # Reap watchers stranded by a deleted project tree before starting a new one.
+  reap_stale_watchers
+
   if [ -z "$MEMSEARCH_CMD" ]; then
     return 0
   fi
@@ -356,5 +449,7 @@ start_watch() {
   else
     $launch_prefix $MEMSEARCH_CMD watch "$MEMORY_DIR" ${COLLECTION_DESC:+--description "$COLLECTION_DESC"} </dev/null &>/dev/null &
   fi
-  echo $! > "$WATCH_PIDFILE"
+  local _watch_pid=$!
+  echo "$_watch_pid" > "$WATCH_PIDFILE"
+  _watch_registry_record "$_watch_pid" "$MEMORY_DIR"
 }
--- a/tests/test_claude_hooks.py
+++ b/tests/test_claude_hooks.py
@@ -1213,3 +1213,132 @@ exit 0
     # No latest version known, so no hint is claimed either way.
     for run in (first, second):
         assert "UPDATE:" not in json.loads(run.stdout)["systemMessage"]
+
+
+COMMON_SH_PLUGINS = ["plugins/claude-code/hooks/common.sh", "plugins/codex/hooks/common.sh"]
+
+
+def _spawn_fake_watcher(tmp_path: Path, target: Path) -> subprocess.Popen[bytes]:
+    """Start a stub standing in for `memsearch watch <target>`.
+
+    Its argv contains "memsearch" so the reaper's PID-reuse guard recognises it,
+    and start_new_session mirrors the setsid/nohup launch in start_watch, which
+    makes it a process-group leader so _kill_tree's group kill has a group.
+    """
+    stub_dir = tmp_path / "stub"
+    stub_dir.mkdir(exist_ok=True)
+    stub = stub_dir / "memsearch"
+    _write_executable(stub, "#!/usr/bin/env bash\nsleep 300\n")
+    return subprocess.Popen([str(stub), "watch", str(target)], start_new_session=True)
+
+
+def _run_registry_snippet(
+    common_sh: str, snippet: str, env: dict[str, str], *args: str
+) -> subprocess.CompletedProcess[str]:
+    return subprocess.run(
+        ["bash", "-c", f'source "$1" < /dev/null\n{snippet}', "bash", common_sh, *args],
+        capture_output=True,
+        text=True,
+        env=env,
+        check=True,
+    )
+
+
+def _registry_env(tmp_path: Path, project: Path, state_dir: Path) -> dict[str, str]:
+    home = tmp_path / "home"
+    home.mkdir(exist_ok=True)
+    return {
+        **os.environ,
+        "HOME": str(home),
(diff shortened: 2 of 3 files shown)


plugins/claude-code/hooks/common.sh
```diff
@@ -289,32 +289,121 @@ kill_orphaned_index() {
...
+# Best-effort guard against PID reuse: a recorded PID may have been recycled by
+# an unrelated process before we get to signal it. Succeeds when the process
+# cannot be inspected at all, which is the unconditional behaviour this
+# replaces rather than a regression.
+_watch_pid_is_ours() {
+  local pid="$1" cmd=""
+  cmd="$(ps -o args= -p "$pid" 2>/dev/null || true)"
+  case "$cmd" in
+    "") return 0 ;;
+    *memsearch*) return 0 ;;
```
yhmo, line 319: plugins/claude-code/hooks/common.sh line:319
Medium ---- `_watch_pid_is_ours` accepts any argv containing the substring "memsearch" as ours, and on a hit `_kill_tree` signals the entire process group (`kill -- -$pid`). If a stale registry entry's PID was recycled to an unrelated process whose argv merely contains "memsearch" (e.g. tooling under a `memsearch*` path), a false positive escalates from one process to its whole group. The guard is documented as best-effort, but the group blast radius is worth tightening: match the `memsearch watch` command form (e.g. `*memsearch watch *`), which still covers the uvx wrapper argv, and/or only group-kill when the group leader's command is confirmed.

plugins/claude-code/hooks/common.sh
```diff
@@ -289,32 +289,121 @@ kill_orphaned_index() {
...
+_watch_registry_forget() {
+  rm -f "$WATCH_REGISTRY_DIR/$1.pid" 2>/dev/null || true
+}
+
+# Reap watchers whose target directory no longer exists. This is the
+# removed-worktree case, which neither the in-tree pidfile (deleted with the
+# tree) nor a "$MEMORY_DIR"-scoped pgrep (no live session ever has that
+# MEMORY_DIR again) can reach. Uses no pgrep, so it also covers platforms where
+# pgrep is unavailable.
+reap_stale_watchers() {
```
yhmo, line 339: plugins/claude-code/hooks/common.sh line:339
Low ---- Only watchers registered from this version onward are reachable by this reaper: pre-fix watchers have no registry entry, and their in-tree pidfile plus the MEMORY_DIR-scoped pgrep sweep are gone with the deleted tree. The 34 orphaned trees / ~1.7 GB cited as motivation are therefore not cleaned up by this change (that measurement came from manual reaping); the fix prevents future leaks only. A one-time migration sweep (e.g. global `pgrep -f "memsearch watch "` with a target-dir liveness check) would close that gap if cleaning existing orphans is intended.

tests/test_claude_hooks.py
```diff
@@ -1213,3 +1213,132 @@ def test_claude_session_start_caches_failed_pypi_lookup(tmp_path: Path) -> None:
...
+            _registry_env(tmp_path, project, state_dir),
+            str(watcher.pid),
+            str(memory),
+        )
+        entry = state_dir / "watchers" / f"{watcher.pid}.pid"
+        assert entry.exists(), "recording a watcher must create its registry entry"
+
+        shutil.rmtree(project)  # the worktree goes away, taking the in-tree pidfile with it
+
+        _run_registry_snippet(common_sh, "reap_stale_watchers", _registry_env(tmp_path, project, state_dir))
```
yhmo, line 1304: tests/test_claude_hooks.py line:1304
Low ---- Coverage gap: these tests exercise `reap_stale_watchers` and the registry path invariant, but nothing drives `stop_watch` with the registry active — `_stop_registered_watchers_for_memory_dir` (stop_watch step 2) and `_watch_registry_forget` (step 1) are untested, and the PID-reuse guard branches (empty ps -> treated as ours; non-memsearch argv -> not ours) are never exercised against a fake non-memsearch process. A test that records a watcher for the current MEMORY_DIR, calls `stop_watch` with the target still present, and asserts the entry is cleared without killing an unrelated pid would cover the main untested surface.

plugins/claude-code/hooks/common.sh
```diff
@@ -289,32 +289,121 @@ kill_orphaned_index() {
...
+      if _watch_pid_is_ours "$pid"; then
+        _kill_tree "$pid"
+      fi
+      rm -f "$entry" 2>/dev/null || true
+    fi
+  done
+}
+
+# Kill any registered watcher whose target is this MEMORY_DIR.
+_stop_registered_watchers_for_memory_dir() {
```
yhmo, line 364: plugins/claude-code/hooks/common.sh line:364
Low ---- `reap_stale_watchers` and `_stop_registered_watchers_for_memory_dir` are near-duplicate registry sweeps (~30 lines): same glob, same `sed -n 1p/2p` parsing, same empty-pid cleanup, differing only in the kill predicate. A future fix to the parsing or cleanup in one must be mirrored in the other manually, and both must also stay in sync with the identical copy in plugins/codex/hooks/common.sh. Extracting a shared per-entry reader (or a predicate-driven sweep) would remove the drift risk.

plugins/claude-code/hooks/common.sh
```diff
@@ -289,32 +289,121 @@ kill_orphaned_index() {
...
   if [ "${MEMSEARCH_NO_WATCH:-}" = "1" ]; then
     return 0
   fi
-  # 1. Kill the process recorded in pidfile
+  # 1. Kill the process recorded in the legacy in-tree pidfile
   if [ -f "$WATCH_PIDFILE" ]; then
     local pid
-    pid=$(cat "$WATCH_PIDFILE" 2>/dev/null)
+    pid="$(cat "$WATCH_PIDFILE" 2>/dev/null || true)"
     if [ -n "$pid" ] && kill -0 "$pid" 2>/dev/null; then
```
yhmo, line 394: plugins/claude-code/hooks/common.sh line:394
Low ---- The retained in-tree pidfile path still kills unconditionally — `_kill_tree "$pid"` runs whenever `kill -0` succeeds, with no `_watch_pid_is_ours` check — even though this PR added that best-effort PID-reuse guard to the new registry paths. This kill site executes at every stop_watch (session end and every start_watch), so a stale pidfile whose PID was recycled onto an unrelated live process is signalled here more often than in reap_stale_watchers. Consider applying the same guard before `_kill_tree`, or documenting why the legacy path stays unguarded. If adopted, the identical code in plugins/codex/hooks/common.sh line:398 needs the same change.

tests/test_claude_hooks.py
```diff
@@ -1213,3 +1213,132 @@ def test_claude_session_start_caches_failed_pypi_lookup(tmp_path: Path) -> None:
...
+    state_dir = tmp_path / "state"
+
+    result = _run_registry_snippet(
+        common_sh,
+        'printf "%s\\n%s\\n" "$WATCH_REGISTRY_DIR" "$MEMSEARCH_DIR"',
+        _registry_env(tmp_path, project, state_dir),
+    )
+    registry_dir, memsearch_dir = result.stdout.splitlines()
+
+    assert Path(registry_dir) == state_dir / "watchers"
```
yhmo, line 1277: tests/test_claude_hooks.py line:1277
Low ---- This test only exercises the explicit `MEMSEARCH_STATE_DIR` override, so the default registry resolution `${XDG_STATE_HOME:-$HOME/.local/state}/memsearch/watchers` that real users hit is never tested. Also, because the hooks prefer `git rev-parse --show-toplevel` when resolving `_PROJECT_DIR`, and pytest runs from the repository root, `$MEMSEARCH_DIR` inside the sourced script resolves to the repo's own `.memsearch` rather than `project/.memsearch`; the two `is_relative_to` assertions then hold trivially and would not catch a regression that placed the default registry inside the watched tree. Consider an additional assertion for the default path (unset `MEMSEARCH_STATE_DIR`, fake `HOME`/`XDG_STATE_HOME`), which would bind the invariant to the actual project tree.

plugins/claude-code/hooks/common.sh
```diff
@@ -289,32 +289,121 @@ kill_orphaned_index() {
...
+}
+
+_watch_registry_forget() {
+  rm -f "$WATCH_REGISTRY_DIR/$1.pid" 2>/dev/null || true
+}
+
+# Reap watchers whose target directory no longer exists. This is the
+# removed-worktree case, which neither the in-tree pidfile (deleted with the
+# tree) nor a "$MEMORY_DIR"-scoped pgrep (no live session ever has that
+# MEMORY_DIR again) can reach. Uses no pgrep, so it also covers platforms where
```
yhmo, line 337: plugins/claude-code/hooks/common.sh line:337
Low ---- The PR description says this reaper also covers #600, but #600 is about `kill_orphaned_index` failing to reap native `memsearch index` processes on Windows — an index-process leak, not a watcher leak. This change only ever reaps *registered watchers* (`reap_stale_watchers`/`_stop_registered_watchers_for_memory_dir`); `kill_orphaned_index` and the index process path are untouched. Suggest clarifying the description/issue linkage so #600 is not closed as fixed by this change (the registry does help the watch half of #658's pgrep blind spot, but not the index half of #600).

zc277584121 (changes requested):
Moving the watcher handle outside the watched tree is the right direction, and the deleted-worktree failure reproduces on the exact base. I cannot approve this head yet because the new reaper can signal processes whose ownership it has not established, and it can lose the new registry handle under concurrency.

The blocking issues are:

1. `_watch_pid_is_ours` succeeds when `ps` returns no command and accepts any argv containing `memsearch`. A recycled PID for an uninspectable process, an unrelated executable under a `memsearch-*` path, or a different `memsearch watch` target is therefore treated as owned. `_kill_tree` then attempts to signal the whole process group without first verifying that the recorded PID is the expected group leader. The legacy in-tree pidfile path still has no identity check at all. These paths must fail closed and validate a positive PID, the `memsearch watch` command form, the recorded target, and the process-group boundary before signalling.

2. Registry records do not carry enough identity to distinguish a removed target from a path that was deleted and recreated. The stale watcher remains indefinitely after path reuse, while PID reuse can turn the same record into a signal against a different process. Please record and validate a stable process identity and target identity, not only PID plus path text.

3. `_watch_registry_record` writes directly to the final file. A deterministic start/stop interleaving can expose the truncated-but-not-yet-written file; the stop-side sweep removes it as empty, then the writer finishes into an unlinked inode. The watcher is left running with no registry handle. Please use an atomic publication protocol and serialize the lifecycle operations, with start/start and start/stop regression coverage.

4. Existing leaked watchers whose worktree and legacy pidfile are already gone have no migration path in this change. That can be an explicit non-goal, but the PR description must say that this prevents future deleted-worktree leaks rather than recovering the already leaked population. Please also remove the claim that this fixes #600: the index cleanup path is untouched. The existing `pgrep` limitations in #658 and the Lite/index behavior tracked by #600 and #629 remain separate.

The current tests pass, but they cover only state-dir override, removed target, and existing target. Please add both-plugin coverage for the default HOME/XDG paths, exited PIDs, uninspectable and recycled PIDs, wrong watch targets, path reuse, current-target stop, legacy pidfiles, atomic/concurrent start and stop, and group-boundary behavior. Non-Linux behavior should be tested with explicit capability stubs unless it is run on native hosts.

PR #706, opened 2026-08-27
feat: add WorkBuddy plugin with memsearch memory hooks

KagaJiankui:
## Summary
Adds a **WorkBuddy (CodeBuddy)** plugin — hook-driven semantic memory built on the `memsearch` CLI. Captures finished turns, summarizes them into daily Markdown journals, indexes into Milvus, and injects the top-2 related memories back into context on prompt submit. Due maintenance (PROJECT.md/USER.md + skill distillation) runs via the shared backend engine through the existing `openclaw` config slot.

## What it does
- **SessionStart** — status line, recent-memory preview, detached one-shot index (skip-if-running), opt-in watch daemon (server mode only), lag-window hint.
- **Stop / SubagentStop** — parse last turn → `memsearch summarize` → append `memory/YYYY-MM-DD.md` (lazy session heading + anchor) → index → write `.last-index-completed` → fire detached due maintenance.
- **UserPromptSubmit** — `memsearch search -k 2 -j` against the project collection; inject hits as `additionalContext`; falls back to a non-blocking marker on any failure.
- **Maintenance** — `handler.py --maintenance` delegates to `memsearch.maintenance.run_due_tasks` + `memsearch.skills.distill`. Config routes through the `openclaw` slot (the backend platform whitelist has no `workbuddy`); prompts resolve to the shared `_shared/prompts`.

## Design notes
- **Thin compat layer** — all heavy lifting (summarize routing, embedding, indexing) is delegated to the CLI; the plugin is stdlib-only Python + JSON hook wiring.
- **Never blocks the session** — every failure path returns `{"continue": true}`.
- Per-project collection derivation matches the existing `derive-collection.sh` convention.

## Tests
- `scripts/tests/` — 26 pytest cases (watch lifecycle, prompt-search injection, maintenance routing) using subprocess doubles / mocked backend; opt-in backend E2E skipped by default.

## Notes
- New plugin dir: `plugins/workbuddy/`. No changes to `plugins/_shared` (prompts are byte-identical).

### DIFF
--- a/plugins/workbuddy/.codebuddy-plugin/plugin.json
+++ b/plugins/workbuddy/.codebuddy-plugin/plugin.json
@@ -1,0 +1,11 @@
+{
+  "name": "memsearch-workbuddy",
+  "version": "0.1.0",
+  "description": "Automatic semantic memory for WorkBuddy — turn capture → daily Markdown journals → Milvus index → top-2 memory injection on prompt submit, with due maintenance (PROJECT.md/USER.md, skill distillation) via the memsearch backend.",
+  "homepage": {
+    "url": "https://github.com/zilliztech/memsearch",
+    "type": "github"
+  },
+  "license": "MIT",
+  "keywords": ["memory", "semantic-search", "milvus", "markdown", "workbuddy"]
+}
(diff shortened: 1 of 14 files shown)


KagaJiankui:
Moved to standalone implementation KagaJiankui/memsearch-workbuddy.

PR #714, opened 2026-09-01
feat: add ZCode platform plugin

karthick-kk:
## Summary

Adds native memsearch support for [ZCode](https://github.com/zilliztech/zcode), a terminal coding agent with a plugin marketplace and shell hook system. ZCode is the sixth supported platform alongside Claude Code, Codex, DeepSeek Harness, OpenClaw, and OpenCode.

ZCode uses the same Stop-hook capture pattern as Claude Code and Codex, but stores conversations in a **SQLite database** (`~/.zcode/cli/db/db.sqlite`) instead of JSONL transcripts, requiring a custom `parse-session.py` to extract turns.

## What ZCode provides for hooks

ZCode fires hooks on 3 lifecycle events (no `SessionEnd`, like Codex):

| Hook | Async | Timeout | Purpose |
|------|-------|---------|---------|
| `SessionStart` | no | 10s | Orphan cleanup, bootstrap memsearch, watch/index, cold-start injection |
| `UserPromptSubmit` | no | 15s | Memory availability hint |
| `Stop` | yes | 120s | Parse last turn from SQLite DB, summarize via `claude -p`, save to daily `.md` |

Key env vars injected by ZCode:
- `CLAUDE_SESSION_ID` — session ID (template variable)
- `ZCODE_PROJECT_DIR` / `CLAUDE_PROJECT_DIR` — project working directory
- `CLAUDE_PLUGIN_ROOT` — plugin install directory (resolves `${CLAUDE_PLUGIN_ROOT}` in `hooks.json`)

## Architecture

The ZCode plugin closely follows the Codex plugin (shell hooks, no SessionEnd, main-context skill), with one key difference: **session storage is SQLite, not JSONL**. The `parse-session.py` script queries the `message` and `part` tables, groups messages into turns (user prompt → assistant response), renders text and tool-call parts, and omits reasoning blocks.

Summarization uses `claude -p --model haiku` (same as the Claude Code plugin) with recursion prevention via `CLAUDECODE=` and `MEMSEARCH_IN_STOP_WORKER=1`. Falls back to truncated raw text if `claude -p` is unavailable.

Memory anchors use the format `<!-- session:<id> db:<path> -->` (instead of `rollout:`/`transcript:`), enabling the memory-recall skill to drill back into the SQLite session database.

### New: `plugins/zcode/`
- `.claude-plugin/plugin.json` — plugin manifest
- `hooks/hooks.json` — declares 3 hooks using `${CLAUDE_PLUGIN_ROOT}` path resolution
- `hooks/common.sh` — env var resolution, JSON helpers, orphan cleanup, Milvus Lite lock handling
- `hooks/stop.sh` — async capture: SQLite session parsing → `claude -p` summarization → daily `.md` append
- `hooks/session-start.sh` — bootstrap, watch/index, cold-start injection
- `hooks/user-prompt-submit.sh` — memory availability hint
- `scripts/parse-session.py` — ZCode SQLite session parser
- `scripts/derive-collection.sh`, `maintenance-runner.py` — synced from `_shared`
- `prompts/`, `skills/` — synced from `_shared`

### Modified: core library
- `src/memsearch/config.py` — added `zcode` to `PluginsConfig` + key mappings
- `src/memsearch/cli.py` — added `zcode` to `--plugin` help, init wizard, summarize prompts
- `plugins/_shared/scripts/maintenance-runner.py` — added `zcode` to `DEFAULT_NATIVE_MODELS`

### Modified: marketplace + shared sources
- `.claude-plugin/marketplace.json` — added `memsearch-zcode` entry
- `plugins/_shared/skills/` — added zcode to platform lists + `references/zcode.md`
- `scripts/sync-skills.sh`, `sync-prompts.sh` — added zcode to sync arrays

### Tests
- `tests/test_zcode_parse_session.py` — 4 tests (last turn, specific turn, nonexistent session, reasoning omission)
- `tests/test_zcode_stop_hook.py` — 3 tests (UTF-8 fallback, empty content skip, no session id)
- `tests/test_config.py` — zcode assertions in set/get roundtrip
- `tests/test_maintenance_runner.py` — zcode added to platform list
- `tests/test_skills_sync.py` — zcode added to `FS_PLATFORMS`

### Docs
- `docs/platforms/zcode/` — index, installation, how-it-works, memory-recall
- `docs/platforms/index.md` — added zcode to comparison table and cross-platform sharing section

## Test results

```
365 passed, 7 skipped, 3 warnings in 48.98s
```

All existing tests pass, plus 7 new zcode-specific tests. Ruff is clean on all new/modified Python files (the 16 pre-existing ruff warnings in `maintenance-runner.py` and `claude-code/transcript.py` are unchanged).

## Checklist

- [x] Plugin follows the same architecture as existing shell-hook platforms (Codex)
- [x] `parse-session.py` handles the SQLite session storage unique to ZCode
- [x] Skills synced from `_shared` (not hand-edited in platform copies)
- [x] Tests cover parse-session.py and stop.sh edge cases
- [x] Docs follow the same structure as other platform docs
- [x] All tests pass (`uv run python -m pytest`)
- [x] Ruff passes on new/modified files
- [x] Conventional commit message used


plugins/zcode/scripts/maintenance-runner.py
```diff
@@ -0,0 +1,593 @@
...
+        env["MEMSEARCH_DISABLE"] = "1"
+        out = run_command(cmd, env=env, cwd=ctx.project_dir, timeout=180)
+        return extract_task_json_output(out)
+
+    raise RuntimeError(f"Unsupported native maintenance platform {ctx.platform!r}")
+
+
+def main() -> int:
+    parser = argparse.ArgumentParser(description="Run plugin-local MemSearch maintenance tasks.")
+    parser.add_argument("--platform", required=True, choices=["claude-code", "codex", "opencode", "openclaw", "dsh"])
```
yhmo: plugins/zcode/scripts/maintenance-runner.py line:526
High ---- The argparse choices for --platform omit "zcode", but common.sh:157 always invokes this runner with --platform zcode, so run_maintenance() fails on every Stop hook (exit code 2, silently swallowed by `|| true`) and the documented PROJECT.md / USER.md / skill-distillation maintenance can never run for the zcode plugin. Even if "zcode" were added to the choices, run_native_provider() has no zcode branch and would raise "Unsupported native maintenance platform". Please add "zcode" to the choices list and mirror the claude-code native branch (claude -p with --safe-mode / --strict-mcp-config / --no-session-persistence / --no-chrome).

plugins/zcode/hooks/session-start.sh
```diff
@@ -0,0 +1,71 @@
...
+    KEY_MISSING=true
+  fi
+fi
+
+start_watch
+
+if [[ "$MILVUS_URI" != http* ]] && [[ "$MILVUS_URI" != tcp* ]]; then
+  ensure_memory_dir
+  if memsearch_available; then
+    _memsearch index "$MEMORY_DIR" >/dev/null 2>&1 || true
```
yhmo: plugins/zcode/hooks/session-start.sh line:61
Medium ---- The Lite-mode one-time index runs synchronously inside the SessionStart hook, but hooks.json gives this hook a 10s timeout and ONNX model warm-up alone can take ~10s, so a cold-start index is likely to be killed mid-run. The sibling claude-code plugin runs this index in a background subshell, records the PID in $INDEX_PIDFILE, and also auto-resets + re-indexes on "dimension mismatch" (plugins/claude-code/hooks/session-start.sh:244-263). Your own docs (docs/platforms/zcode/how-it-works.md) describe "a single background index run" and "dimension mismatch auto-recovery" that this code does not implement. Please background the index (recording the PID) and add the dimension-mismatch recovery so the code matches its docs and survives the hook timeout.

plugins/zcode/hooks/session-start.sh
```diff
@@ -0,0 +1,71 @@
...
+  fi
+fi
+
+ADDITIONAL_CONTEXT=""
```
yhmo: plugins/zcode/hooks/session-start.sh line:65
Medium ---- ADDITIONAL_CONTEXT is declared but never used, and the documented cold-start features are absent: docs/platforms/zcode/how-it-works.md says the SessionStart hook "injects memory file count and date range, with a hint to use /memory-recall" and performs an "update check" against PyPI, but the script only returns '{}'. Additionally, index_state_warning() and skill_candidate_hint() print raw WARNING/hint text to stdout before the trailing echo '{}', so when either triggers the hook emits non-JSON output (the sibling claude-code hook captures these into a JSON systemMessage, plugins/claude-code/hooks/session-start.sh:117,265-272). Either implement the documented injection/update-check and wrap warnings in a systemMessage, or update the docs to describe the actual behavior.

plugins/zcode/scripts/parse-session.py
```diff
@@ -0,0 +1,200 @@
...
+def _connect(db_path: str) -> sqlite3.Connection:
+    if not Path(db_path).is_file():
+        raise RuntimeError(f"ZCode DB not found: {db_path}")
+    uri = f"file:{db_path}?mode=ro"
+    return sqlite3.connect(uri, uri=True)
+
+
+def _load_messages(conn: sqlite3.Connection, session_id: str) -> list[dict]:
+    rows = conn.execute(
+        "SELECT id, sequence, data FROM message WHERE session_id = ? ORDER BY sequence, time_created, id",
```
yhmo, line 47: plugins/zcode/scripts/parse-session.py line:47
Medium ---- The entire capture path depends on an assumed ZCode SQLite schema (message/part tables, time_created, data JSON with semantics.kind / anchor.turnId, part type text/tool/reasoning), but zilliztech/zcode is not publicly accessible and the tests in tests/test_zcode_parse_session.py build that schema from the same assumption, so they prove self-consistency rather than compatibility with a real ZCode database. Please confirm the schema — in particular the time_created column used in the ORDER BY here and at line 70 — was validated against a real ZCode install (for example, a schema dump or a fixture captured from an actual session), since a wrong column or part type would silently disable capture for every user.

docs/platforms/zcode/how-it-works.md
```diff
@@ -0,0 +1,299 @@
...
+├── 2026-03-25.md
+└── 2026-03-26.md
+```
+
+### Example Memory File
+
+```markdown
+# 2026-03-25
+
+## Session 10:30
```
yhmo, line 230: docs/platforms/zcode/how-it-works.md line:230
Low ---- The example memory file shows a "## Session 10:30" heading, but stop.sh only writes "### HH:MM" entries and never emits a "## Session" heading, so the documented example does not match the actual capture format. The memory-recall skill also greps for "^## " to list session headings across days, which would return nothing for zcode-written journals. Please align the example (and the recall skill's session-heading assumption) with the real output format.

plugins/_shared/skills/memory-config/SKILL.md
```diff
@@ -25,11 +25,12 @@ matching your current environment:
 - OpenClaw → `references/openclaw.md`
 - OpenCode → `references/opencode.md`
 - DeepSeek Harness → `references/dsh.md`
+- ZCode → `references/zcode.md`
 
 If you are unsure which agent you are, check these environment markers:
 `DSH_HOME`/`~/.dsh` → DeepSeek Harness; `CODEX_HOME`/`~/.codex` → Codex;
 `~/.openclaw` → OpenClaw; `~/.config/opencode` → OpenCode;
-`CLAUDE_PLUGIN_ROOT` → Claude Code.
+`CLAUDE_PLUGIN_ROOT` → Claude Code; `~/.zcode` → ZCode.
```
yhmo: plugins/_shared/skills/memory-config/SKILL.md line:33
Low ---- The environment-marker guidance is ambiguous for ZCode: ZCode also exports CLAUDE_PLUGIN_ROOT (it is used to resolve ${CLAUDE_PLUGIN_ROOT} in hooks/hooks.json), so a ZCode user's environment matches BOTH the "CLAUDE_PLUGIN_ROOT -> Claude Code" marker and the "~/.zcode -> ZCode" marker. An agent matching markers in the listed order could misidentify the platform as Claude Code and read the wrong reference file. Please document precedence (for example, check ~/.zcode first for ZCode) in this line and in the synced platform copies.

plugins/_shared/skills/memory-config/references/zcode.md
```diff
@@ -0,0 +1,72 @@
...
+entries) to find the plugin source path, then compare that repository with the
+latest `zilliztech/memsearch` GitHub release:
+
+```bash
+git -C <memsearch-repo> describe --tags --always --dirty
+gh release view --repo zilliztech/memsearch --json tagName,publishedAt,url
+```
+
+Update source installs with `git pull` plus
+`bash plugins/zcode/scripts/install.sh`.
```
yhmo: plugins/_shared/skills/memory-config/references/zcode.md line:16
Low ---- This reference instructs source-install users to run "bash plugins/zcode/scripts/install.sh", but plugins/zcode/scripts/ contains only derive-collection.sh, maintenance-runner.py, and parse-session.py — no install.sh exists, so the documented update step cannot work. Please either ship the installer or point source installs at the marketplace/source install flow described in docs/platforms/zcode/installation.md.

karthick-kk:
Thanks for the review! Fixed in commit `b405236`:

- Added `zcode` to the `--platform` choices
- Added a `zcode` branch to `run_native_provider()` that mirrors the `claude-code` branch (uses `claude -p` with the same flags)

The plugin now works for PROJECT.md / USER.md / skill-distillation maintenance tasks.

karthick-kk:
Fixed in commit `b405236`:

- The Lite-mode index now runs in a background subshell (`&`) to avoid the 10s hook timeout
- Added `kill_orphaned_index` call before launching to clean up stale PIDs  
- Added dimension mismatch auto-recovery (reset + re-index)
- Added PID file write (`echo $! > "$INDEX_PIDFILE"`) so `kill_orphaned_index` can find and kill stale index processes in subsequent sessions

This matches the pattern in `plugins/claude-code/hooks/session-start.sh:244-262`.

karthick-kk:
Fixed in commit `b405236`:

- Added `_resolve_symlinks`, `_installed_version_from_dist_info`, and `_pypi_latest_version` to `plugins/zcode/hooks/common.sh` (were called but not defined)
- SessionStart now collects: memory file count + dates, index/skill warnings, PyPI update check
- All output wrapped in a single JSON `systemMessage` to ensure valid JSON output
- Added session heading creation (`## Session HH:MM`) to match codex plugin and the documented memory format

This fixes the "ADDITIONAL_CONTEXT declared but never used" and the raw text output issues.

karthick-kk:
This is a valid concern:  the SQLite schema assumptions in `parse-session.py` come from inspecting the ZCode source code and the existing bridge plugin (`plugins/local/zcode-memsearch-bridge/hooks/parse-transcript.py`).

The schema assumptions:
- `message` table with `id`, `sequence`, `session_id`, `data` (JSON), `time_created`
- `part` table with `message_id`, `data` (JSON), `type`, `sequence`, `time_created`
- `data` JSON fields: `role`, `semantics.kind` (`user_prompt`/`assistant_response`), `anchor.turnId`
- `part.type`: `text`, `tool`, `reasoning`

These match what the existing bridge plugin uses. I added a comment to `parse-session.py` documenting this dependency.

If you have access to a ZCode instance, you can dump the schema with:
```bash
sqlite3 ~/.zcode/cli/db/db.sqlite ".schema"
```

We can refine the parser once we have a confirmed schema from a real ZCode install.

karthick-kk:
Fixed in commit `d4fdef7`:

- SessionStart now writes `## Session $NOW` heading to the daily memory file before any entries
- This matches the codex plugin pattern (`plugins/codex/hooks/session-start.sh:125-126`)  
- The memory-recall skill now finds session headings when grepping `^## `

The example in `docs/platforms/zcode/how-it-works.md` will now match the actual output.

karthick-kk:
Fixed in commit `d4fdef7`:

- Changed the marker list order in `plugins/_shared/skills/memory-config/SKILL.md` to check `~/.zcode` first
- ZCode also exports `CLAUDE_PLUGIN_ROOT`, so it was matching the Claude Code marker before

Updated order:
```
~/.zcode → ZCode; DSH_HOME/~/.dsh → DeepSeek Harness; ...
CLAUDE_PLUGIN_ROOT → Claude Code.
```

karthick-kk:
Fixed in commit `d4fdef7`:

- Removed the reference to the non-existent `plugins/zcode/scripts/install.sh`
- Updated to describe the actual ZCode installation flow: marketplace install + `git pull` for source updates

ZCode has no `install.sh` because it installs plugins via its marketplace system, not a standalone script.

plugins/zcode/scripts/maintenance-runner.py
karthick-kk: Fixed in commit b405236:

- Added `zcode` to the `--platform` choices
- Added a `zcode` branch to `run_native_provider()` mirroring the claude-code branch (uses `claude -p` with the same flags)

Now maintenance tasks (PROJECT.md, USER.md, skill distillation) work for the zcode plugin.

plugins/zcode/hooks/session-start.sh
karthick-kk: Fixed in commit b405236:

- Lite-mode index now runs in a background subshell (`&`) to avoid the 10s hook timeout
- Added `kill_orphaned_index` call before launching
- Added dimension mismatch auto-recovery (reset + re-index)
- Added PID file write (`echo $! > "$INDEX_PIDFILE"`) so orphan cleanup works in subsequent sessions

Matches the pattern in `plugins/claude-code/hooks/session-start.sh:244-262`.

plugins/zcode/hooks/session-start.sh
karthick-kk: Fixed in commit b405236:

- Added `_resolve_symlinks`, `_installed_version_from_dist_info`, `_pypi_latest_version` to zcode/common.sh (were called but not defined)
- SessionStart now collects: memory file count + dates, index/skill warnings, PyPI update check
- All output wrapped in a single JSON `systemMessage` to ensure valid JSON output
- Added session heading creation (`## Session HH:MM`) to match codex pattern

This fixes "ADDITIONAL_CONTEXT declared but never used" and the raw text output issues.

plugins/zcode/scripts/parse-session.py
karthick-kk: Valid concern — the schema assumptions come from inspecting ZCode source code and the existing bridge plugin (`plugins/local/zcode-memsearch-bridge/hooks/parse-transcript.py`).

Assumptions:
- `message` table: `id`, `sequence`, `session_id`, `data` (JSON), `time_created`
- `part` table: `message_id`, `data` (JSON), `type`, `sequence`, `time_created`
- `data` JSON: `role`, `semantics.kind` (`user_prompt`/`assistant_response`), `anchor.turnId`
- `part.type`: `text`, `tool`, `reasoning`

These match what the bridge uses. Added a comment in parse-session.py documenting this dependency. We can refine the parser once we have a confirmed schema from a real ZCode install.

docs/platforms/zcode/how-it-works.md
karthick-kk: Fixed in commit d4fdef7:

- SessionStart now writes `## Session $NOW` heading before any entries
- Matches the codex plugin pattern (`plugins/codex/hooks/session-start.sh:125-126`)
- The memory-recall skill now finds session headings when grepping `^## `

The example in how-it-works.md now matches the actual output.

plugins/_shared/skills/memory-config/SKILL.md
karthick-kk: Fixed in commit d4fdef7:

- Changed marker list order to check `~/.zcode` first
- ZCode also exports `CLAUDE_PLUGIN_ROOT`, so it was matching the Claude Code marker before

Updated order:
```
~/.zcode → ZCode; DSH_HOME/~/.dsh → DeepSeek Harness; ...
CLAUDE_PLUGIN_ROOT → Claude Code.
```

plugins/_shared/skills/memory-config/references/zcode.md
karthick-kk: Fixed in commit d4fdef7:

- Removed reference to non-existent `plugins/zcode/scripts/install.sh`
- Updated to describe the actual ZCode installation flow: marketplace install + `git pull` for source updates

ZCode has no install.sh because it installs plugins via its marketplace system.

PR #718, opened 2026-09-04
feat: add Voyage AI reranker provider

georgisokolov:
## Problem

Reranking currently requires a local cross-encoder. That has two costs for the
short-lived processes that dominate real usage (the plugin hooks and the
`memory-recall` skill each shell out to `memsearch search`):

1. **Per-process model load.** ~150MB of weights are loaded from disk on every
   invocation, since nothing survives between CLI runs.
2. **A hard network dependency.** `_load_onnx_model` calls `list_repo_files(repo_id)`
   on every call, even when the model is fully cached, so reranking with
   `HF_HUB_OFFLINE=1` raises rather than degrading.

Measured on an indexed corpus, enabling the local reranker took search from
~3.5s to ~14s per invocation.

## Change

Adds a hosted Voyage AI rerank backend, selected by a new `reranker.provider`
setting that mirrors how `embedding.provider` selects an embedding backend:

```toml
[reranker]
provider = "voyage"
model = "rerank-3"      # optional; this is the default
```

Also available as `memsearch search --reranker-provider voyage`.

Added latency measured through the CLI on the same corpus and query: **~0.35s
for the hosted provider vs ~10.5s for the local cross-encoder** (the constant
interpreter/`uv run` overhead was held the same across both).

## Notes on the design

- **`provider` mirrors `embedding.provider`** rather than encoding the backend in
  the model string. I first prototyped a prefix form (`voyageai/rerank-3`) and a
  bare-name heuristic, but a `provider` field matches the existing config shape,
  needs no shape-sniffing, and does not shadow the real `voyageai` HuggingFace
  namespace.
- **Backwards compatible.** An empty `provider` keeps the current auto-detected
  local behaviour. `MemSearch` now enables reranking on
  `reranker_model or reranker_provider`, so `provider` alone works and the
  backend supplies its default model.
- **No new dependency.** Reuses the existing `voyage` extra (`memsearch[voyage]`),
  already present for the Voyage embedding provider.
- **Degradation matches the local backends.** A missing `voyageai` package logs a
  warning and returns results unranked, the same as an absent local backend. API
  errors (bad key, network) propagate, as an ONNX/HF failure does today. Happy to
  make those degrade instead if you would prefer that.
- The pre-existing `list_repo_files` per-call network hit is left alone to keep
  this PR focused; happy to open a separate one for it.

## Testing

- New `tests/test_reranker_voyage.py` (8 tests, fake `voyageai` module): API
  dispatch and result mapping, default-model fallback, `top_k` pass-through,
  `top_k=0` meaning all, missing-package degradation, unknown-provider
  degradation, local models never reaching the API, and empty-input
  short-circuit.
- Extended `tests/test_cli_config_helpers.py` for the new config key in both the
  override map and the resolved kwargs.
- Full suite: **386 passed, 7 skipped**. `ruff check` and `ruff format --check`
  both clean.
- Verified end to end against the live Voyage API and a real Milvus collection,
  with the expected chunk returned at rank 1.

### DIFF
--- a/src/memsearch/reranker.py
+++ b/src/memsearch/reranker.py
@@ -4,11 +4,20 @@ Re-scores hybrid search results using a cross-encoder that reads query and
 document together, producing more accurate relevance scores than embedding
 similarity alone.
 
-Two backends are supported, auto-detected at runtime:
-  1. ONNX Runtime (preferred) — lightweight, CPU-only, included in ``memsearch[onnx]``
-  2. sentence-transformers CrossEncoder — PyTorch-based, included in ``memsearch[local]``
+Three backends are supported, chosen by the ``reranker.provider`` setting the
+same way ``embedding.provider`` selects an embedding backend:
+  1. ``"voyage"`` — Voyage AI hosted rerank API; needs ``memsearch[voyage]``
+     and ``VOYAGE_API_KEY``
+  2. ``""`` (default) — a local cross-encoder, auto-detecting the best runtime:
+     ONNX Runtime (``memsearch[onnx]``), else sentence-transformers
+     (``memsearch[local]``)
 
-If neither is installed, reranking is silently skipped.
+Local backends load a cross-encoder from HuggingFace, so the first call downloads
+weights and every process pays the load cost; the hosted backend has no local
+model to load, which keeps per-process search latency low for short-lived CLI
+invocations.
+
+If no backend is available, reranking is silently skipped.
 """
 
 from __future__ import annotations
@@ -229,6 +238,41 @@ def _rerank_torch(query: str, results: list[dict[str, Any]], model_name: str, to
 
 
 # ======================================================================
+# Voyage AI backend (hosted API)
+# ======================================================================
+
+DEFAULT_VOYAGE_RERANKER = "rerank-3"
+
+_voyage_client: Any = None
+_voyage_client_lock = threading.Lock()
+
+
+def _load_voyage_client() -> Any:
+    """Return a cached Voyage client (reads ``VOYAGE_API_KEY``)."""
+    global _voyage_client
+    with _voyage_client_lock:
+        if _voyage_client is None:
+            import voyageai
+
+            _voyage_client = voyageai.Client()
+            logger.info("Initialised Voyage AI reranker client")
+        return _voyage_client
+
+
+def _rerank_voyage(query: str, results: list[dict[str, Any]], model_name: str, top_k: int) -> list[dict[str, Any]]:
+    """Rerank using the Voyage AI rerank endpoint.
+
+    The API returns its results already ordered by relevance, each carrying the
+    index of the document it scored, so the original result dicts are looked up
+    by that index rather than re-sorted locally.
+    """
+    client = _load_voyage_client()
+    documents = [r["content"] for r in results]
+    response = client.rerank(query, documents, model=model_name, top_k=top_k or None)
+    return [{**results[item.index], "score": float(item.relevance_score)} for item in response.results]
+
+
+# ======================================================================
 # Public API
 # ======================================================================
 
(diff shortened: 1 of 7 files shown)


src/memsearch/reranker.py
```diff
@@ -254,19 +295,41 @@ def rerank(
...
 
+    if provider == "voyage":
+        try:
+            return _rerank_voyage(query, results, model_name or DEFAULT_VOYAGE_RERANKER, top_k)
+        except ImportError:
+            logger.warning(
+                'Reranker provider "voyage" requires the voyageai client '
+                '(pip install "memsearch[voyage]"); skipping reranking',
+            )
+            return results
```
yhmo: src/memsearch/reranker.py line:324
Medium ---- When the voyage provider degrades (missing voyageai package, unknown provider, or no local backend), this path returns the full candidate list unchanged, while MemSearch.search() over-fetches top_k * 3 candidates. A user who enables --reranker-provider voyage without the extra gets up to 3x the requested result count (e.g. 30 results for --top-k 10) instead of top_k, violating the top_k contract documented in the rerank() docstring. Suggest truncating on all three degradation returns (here, line 331, and line 343): `return results[:top_k] if top_k > 0 else results`. The new tests call rerank() directly with 3 items and never exercise the top_k * 3 interplay, so the over-delivery is currently uncaught.

src/memsearch/reranker.py
```diff
@@ -254,19 +295,41 @@ def rerank(
...
+        Results re-sorted by relevance score. Returned unchanged when no
+        backend is available.
     """
     if not results:
         return []
 
+    if provider == "voyage":
+        try:
+            return _rerank_voyage(query, results, model_name or DEFAULT_VOYAGE_RERANKER, top_k)
+        except ImportError:
```
yhmo, line 332: src/memsearch/reranker.py line:319
Low ---- Question: this `except ImportError` only covers a missing voyageai package. When VOYAGE_API_KEY is unset, voyageai.Client.rerank() raises voyageai.error.AuthenticationError (verified against the SDK source), which is not an ImportError, so it propagates out of rerank() and MemSearch.search() as a raw traceback (the CLI search command only catches MilvusException). A missing package degrades to unranked results while a missing key crashes, which is asymmetric for the same feature. Would you prefer catching voyageai.error.VoyageError here too, logging a warning and returning unranked results, for consistency with the missing-package path?

src/memsearch/reranker.py
georgisokolov: Fixed in a52c130. All four skip-reranking returns now cap at `top_k` (`results[:top_k] if top_k > 0 else results`), matching what `_rerank_onnx` and `_rerank_torch` already did.

Reproduced it first rather than patching on description: a `search()` test with a fake store returns 30 results for `--top-k 10` when the voyage extra is missing. That test is now in the suite as `test_search_does_not_over_deliver_when_reranking_degrades`, alongside a per-path `top_k` test for each degradation return and one pinning that `top_k=0` still returns every candidate. Each was mutation-checked: reverting any single truncation fails exactly the test that covers it, and nothing else.

Also corrected the `rerank()` docstring, which promised results were "returned unchanged" on that path.

src/memsearch/reranker.py
georgisokolov: Catching it. Confirmed against voyageai 0.5.0 at runtime rather than from the source: `Client()` constructs fine with no key, and `rerank()` then raises `voyageai.error.AuthenticationError`, which is a `VoyageError` and not an `ImportError`.

a52c130 adds an `except _voyage_error_types()` clause beside the `ImportError` branch: it logs a warning naming the error class and returns the unranked results capped at `top_k`, so a missing key now behaves like a missing package. Catching the `VoyageError` base rather than `AuthenticationError` alone also covers `RateLimitError`, `Timeout` and `APIConnectionError`, which have the same answer here: do not fail the search over a reranking step that is an optional refinement.

`_voyage_error_types()` returns an empty tuple when voyageai is absent, and `except ()` never matches, so the clause is safe to name unconditionally without a module-level import. A test pins that a non-Voyage exception still propagates rather than being swallowed by the new handler.

The degradation contract is now documented in `docs/home/configuration.md`.

zc277584121:
Thanks for the thoughtful contribution and for addressing the review feedback. The implementation looks promising, but we need a bit more time to settle the public reranker configuration and backend boundary before deciding how to land it. No further action is needed from you right now; we'll keep the PR open and follow up when we're ready to continue.

src/memsearch/reranker.py
yhmo: The issue is confirmed fixed.

src/memsearch/reranker.py
yhmo: The issue is confirmed fixed.

PR #721, opened 2026-09-06
fix(dsh): support native Windows runtime

andreyt2002-droid:
## Summary

Adds native Windows support for the DeepSeek Harness plugin while preserving the existing POSIX runtime path.

- use direct executable/argument spawning instead of Bash/WSL dependencies;
- resolve executable paths safely on Windows, rejecting `.cmd`/`.bat` wrappers for direct spawn;
- derive DSH collection names in JavaScript with Linux-compatible output;
- prefer the checkout-local `Scripts/python.exe` beside an explicit `MEMSEARCH_CMD`;
- pass agent identity through the environment and keep summarizer stdout/stderr UTF-8-safe;
- add Windows-safe fixtures, ambient-command isolation, and Windows CI coverage;
- add best-effort local JSONL auditing for background LLM calls (`memsearch audit`) and DSH headless/custom-summary failures;
- document `summarizeTimeoutMs` overrides for slow custom LLM providers and synchronize shared skill references.

## Validation

- Windows: `npm test` in `plugins/dsh` — 65 passed, 1 skipped.
- Windows: `uv run python -m pytest plugins/dsh/tests -q` — 25 passed.
- Linux isolated parity worktree: `npm test` — 64 passed, 2 skipped; Python DSH/audit tests — 26 passed.
- Latest GitHub Actions run is green: Ruff, Python 3.10–3.13, OpenCode, and DSH plugin tests on Ubuntu and Windows.

## Notes

The standard summary timeout remains 30 seconds. Slow cloud providers can opt in to a larger value with the DSH profile setting `summarizeTimeoutMs`; the local Web profile uses 120000 ms with `summarizeMode: custom-llm`.

### DIFF
--- a/plugins/dsh/index.js
+++ b/plugins/dsh/index.js
@@ -259,6 +259,22 @@ async function createMemoryMessage(ctx, text) {
 
 /** Detect the memsearch CLI as an executable plus fixed leading arguments. */
 function detectMemsearchCmd() {
+  const explicit = process.env.MEMSEARCH_CMD
+  if (explicit) {
+    if (explicit.trim().startsWith('[')) {
+      let argv
+      try {
+        argv = JSON.parse(explicit)
+      } catch {
+        throw new Error('MEMSEARCH_CMD must be valid JSON when specified as an argv array')
+      }
+      if (!Array.isArray(argv) || argv.length === 0 || argv.some((value) => typeof value !== 'string') || !argv[0]) {
+        throw new Error('MEMSEARCH_CMD must be a nonempty executable argv array')
+      }
+      return argv
+    }
+    return [explicit]
+  }
   const onPath = resolveExecutable('memsearch')
   if (onPath) return [onPath]
   const localUvx = join(runtimeHome(), '.local', 'bin', process.platform === 'win32' ? 'uvx.exe' : 'uvx')
@@ -288,13 +304,9 @@ function deriveCollection(projectDir, override) {
 
 function requireDefaultCollectionSupport(memsearchCmd, projectDir, collection) {
   try {
-    execFileSync(
-      'bash',
-      [
-        '-c',
-        `${memsearchCmd} config get milvus.collection ` +
-          `--default-collection '${shellEscape(collection)}'`,
-      ],
+    execCommandSync(
+      memsearchCmd,
+      ['config', 'get', 'milvus.collection', '--default-collection', collection],
       {
         cwd: projectDir,
         encoding: 'utf-8',
--- a/plugins/dsh/index.js
+++ b/plugins/dsh/index.js
@@ -1081,12 +1081,13 @@ function summarizeCustomLlm(opts, render, projectDir) {
 /**
  * Resolve the dsh CLI command for one-shot headless summarization.
  *
- * `dsh` may not be on PATH (it is a pnpm-installed workspace bin). We check
- * PATH first, then `DSH_CLI` as an explicit override, then the pnpm global
- * bin directory. Returns the command as an argv array (`[cmd, ...args]`),
- * or null when not found. The array form is required because `DSH_CLI` may
- * be an interpreter invocation (e.g. `node /path/to/bin.js`), which `spawn`
- * cannot treat as a single executable.
+ * `dsh` may not be on PATH (it is a pnpm-installed workspace bin). Explicit
+ * environment overrides win first — `MEMSEARCH_DSH_COMMAND_JSON` (a JSON argv
+ * array), then `DSH_CLI` (a quoted shell-style command line) — then executable
+ * discovery on PATH, then the pnpm global bin directory. Returns the command
+ * as an argv array (`[cmd, ...args]`), or null when not found. The array form
+ * is required because the overrides may be an interpreter invocation (e.g.
+ * `node /path/to/bin.js`), which `spawn` cannot treat as a single executable.
  */
 function detectDshCmd() {
   const json = process.env.MEMSEARCH_DSH_COMMAND_JSON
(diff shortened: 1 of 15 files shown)


plugins/dsh/index.js
```diff
@@ -138,47 +256,33 @@ async function createMemoryMessage(ctx, text) {
...
-    return undefined
+    // Nonexistent paths keep their resolved absolute spelling, matching realpath -m.
   }
+  const sanitized = basename(absolute)
+    .toLowerCase()
+    .replace(/[^a-z0-9]/g, '_')
+    .replace(/_+/g, '_')
+    .replace(/^_|_$/g, '')
+    .slice(0, 40)
+  const hash = createHash('sha256').update(absolute).digest('hex').slice(0, 8)
```
yhmo, line 301: plugins/dsh/index.js line:284
Medium ---- deriveCollection() is a new JavaScript port of scripts/derive-collection.sh and is now the DSH plugin's only runtime derivation path, but no test in plugins/dsh/tests/index.test.js exercises it or checks parity with the bash script, and the plugin's other platforms still shell out to derive-collection.sh. The collection name is the join key between indexMemory() and runSearch() across all agents, so a silent divergence would make recall return nothing with no error; today the parity claim rests only on manual WSL validation. Consider a test that runs representative project paths (spaces, quotes, non-ASCII, trailing separators, symlinked parents) through both implementations and asserts identical output, similar to plugins/opencode/derive-collection.test.ts.

plugins/dsh/index.js
```diff
@@ -858,20 +949,39 @@ function summarizeCustomLlm(opts, render, projectDir) {
...
+      // Fall through to the legacy override and executable discovery.
     }
   }
-  if (onPath('dsh')) return ['dsh']
   const fromEnv = process.env.DSH_CLI
-  if (fromEnv) return fromEnv.split(/\s+/).filter(Boolean)
-  const home = process.env.HOME || ''
-  const pnpmBin = join(home, '.local', 'share', 'pnpm')
-  if (existsSync(join(pnpmBin, 'dsh'))) return [join(pnpmBin, 'dsh')]
+  if (fromEnv) {
```
yhmo, line 1000: plugins/dsh/index.js line:962
Low ---- This new code now honors MEMSEARCH_DSH_COMMAND_JSON and DSH_CLI before checking PATH (a precedence change from the previous PATH-first logic), but the doc comment above detectDshCmd() still says "We check PATH first, then DSH_CLI as an explicit override" and tests/index.test.js still claims "DSH_CLI is checked after PATH". Please update both so a future maintainer does not reorder the code back or misjudge DSH_CLI as a mere fallback.

plugins/dsh/index.js
```diff
@@ -986,10 +1096,11 @@ function registerMemoryRecallSkill(ctx, opts, memsearchCmd, collection, projectD
     .replace(/^﻿?---[\s\S]*?---\s*/u, '') // strip optional YAML frontmatter
     .replace(/^<!--[\s\S]*?-->\s*/u, '') // strip the human-facing metadata comment
     .replaceAll('{{AGENT_NAME}}', agentName)
-    .replaceAll('{{MEMSEARCH_CMD}}', memsearchCmd)
+    .replaceAll('{{MEMSEARCH_CMD}}', typeof memsearchCmd === 'string' ? memsearchCmd : commandForSkill(memsearchCmd))
+    .replaceAll('{{PYTHON_CMD}}', commandForSkill(detectPythonCmd() ?? ['python3']))
     .replaceAll('{{PLUGIN_DIR}}', PLUGIN_DIR)
     .replaceAll('{{PROJECT_DIR}}', projectDir)
-    .replaceAll('{{COLLECTION}}', collection || `$(bash "${PLUGIN_DIR}/scripts/derive-collection.sh")`)
+    .replaceAll('{{COLLECTION}}', collection || deriveCollection(PLUGIN_DIR, ''))
```
yhmo: plugins/dsh/index.js line:1103
Low ---- If this {{COLLECTION}} substitution is ever reached with an empty collection, the fallback derives the collection from PLUGIN_DIR (a fixed constant) instead of projectDir, so the memory-recall skill would search one fixed collection regardless of project. It is unreachable today because callers always pass a derived bootCollection, but the wrong constant is a trap for the next caller; consider deriveCollection(projectDir, '') or dropping the fallback.

yhmo (commented):
Overall note: this branch is currently not mergeable (`mergeable: false`, state `dirty`). `main` has advanced past this PR's base (6f86211) after PR #722 merged, touching several of the same files (`plugins/dsh/index.js`, `plugins/dsh/tests/index.test.js`, `plugins/dsh/README.md`, and the `dsh.md` skill references). No check runs have executed on head `2915b7ce`, so the "42/42 passed on Windows" validation in the description is not yet reproducible from CI. Please rebase onto `main` so the new `windows-latest` job can run.

The three inline comments are non-blocking: one Medium suggestion for a parity test of the new JavaScript collection-derivation port, and two Low cleanups (stale `detectDshCmd` precedence docs, and a wrong-constant dead fallback for the memory-recall `{{COLLECTION}}` substitution).

andreyt2002-droid:
Thanks for the review! Addressed everything in 783495c (rebased onto current main):

- **deriveCollection parity**: added a test that runs representative project paths (spaces, quotes, a backslash segment, non-ASCII, trailing separator) through both `scripts/derive-collection.sh` and the JavaScript port, asserting identical output. It runs wherever bash is available (Linux CI); on win32 it is skipped with a reason because the bash script uses POSIX path semantics while the win32 runtime derivation is JavaScript-only. A companion win32 unit test covers sanitization of spaces/non-ASCII basenames, stability, and hash distinctness.
- **detectDshCmd doc/test drift**: the doc comment and the stale test comment now state that `MEMSEARCH_DSH_COMMAND_JSON` and then `DSH_CLI` take precedence over PATH discovery.
- **{{COLLECTION}} fallback**: now derives from `projectDir` instead of `PLUGIN_DIR`.

One note from the rebase: the two new injection tests from #722 route their fake CLI through `/bin/sh` and `/usr/bin/bash` shims, so they cannot run on `windows-latest`. They are skipped on win32 with that reason in this branch to keep the Windows CI job green; a cross-platform fixture for them would make a good follow-up.

plugins/dsh/index.js
yhmo: The issue is confirmed fixed.

plugins/dsh/index.js
yhmo: The issue is confirmed fixed.

plugins/dsh/index.js
yhmo: The issue is confirmed fixed.

andreyt2002-droid:
All CI checks are green on the current `eb0ee81` head, and the PR is mergeable with no static findings. Native Windows parity and the DSH/audit follow-ups have been validated on Windows and in an isolated Linux worktree. The branch is intentionally frozen while awaiting maintainer review. Could you please merge #721 when convenient?

PR #725, opened 2026-09-07
feat(plugins): make the daily memory filename collision-safe on synced directories

### LINKED ISSUE #720: Daily memory filename collides across machines sharing a synced .memsearch/memory/ folder
**Summary**

I run Claude Code + the memsearch plugin on two Macs against the same project repo. The repo's working directory (`~/Documents/<project>`) is synced between the two machines via iCloud Drive (file-level sync, gitignored paths included). This means `.memsearch/memory/` ends up populated by both machines, which is actually great — semantic search on either machine surfaces sessions from both.

The problem is the daily memory filename is fixed to just the date:

```
MEMORY_FILE="$MEMORY_DIR/$TODAY.md"
```

(same line in both `hooks/session-start.sh` and `hooks/stop.sh`). The `Stop` hook fires after *every* turn and appends to this file, so when both machines are active on the same day, they're both appending to the *same physical file* concurrently. A file-level sync provider (iCloud, Dropbox, Syncthing, etc.) doesn't merge concurrent plain-text appends — it either produces a `(conflicted copy)` duplicate or silently resolves via last-writer-wins, in which case one machine's turn summary is quietly dropped from the memory corpus with no error.

**Proposed enhancement**

Make the daily filename collision-safe for multi-writer setups by including a per-writer identifier (e.g. hostname) in the filename, e.g.:

```
MEMORY_FILE="$MEMORY_DIR/$TODAY-$(hostname -s).md"
```

Since `memsearch index "$MEMORY_DIR"` already ingests every `*.md` file in the directory regardless of name, this requires no change to indexing/search — only to the filename the hooks write to. Each machine would own its own file, eliminating the write-write race entirely, while a shared/synced memory directory still gives full cross-machine search.

Ideas for keeping this backward-compatible / opt-in:
- Default to current behavior (`$TODAY.md`, no suffix) so single-machine users see no change.
- Add a config knob (e.g. `memory.filename_suffix = "hostname"` in `~/.memsearch/config.toml`, or an env var like `MEMSEARCH_FILENAME_SUFFIX`) that multi-machine users can opt into.
- Alternatively, always suffix with a short stable per-install ID (already have something like this for collection naming via `derive-collection.sh`) rather than relying on `hostname -s`, in case two machines happen to share a hostname.

**Why this matters**

Nothing catastrophic is lost when this happens (worst case: one turn's summary bullet points don't make it into the search index), but it's a silent, hard-to-notice data-loss mode for anyone running memsearch across multiple machines against a synced project folder — a setup the plugin already implicitly supports well (since indexing just globs the memory dir).

**Environment**
- memsearch 0.4.19 (`uv tool install 'memsearch[onnx]'`)
- memsearch Claude Code plugin 0.4.0
- macOS, two machines, project folder synced via iCloud Drive
- embedding provider: onnx (local)

**Related:** #667 proposes a general filename templating mechanism (`{date}`, `{project}`, `{session}`) for a different use case (multi-project on one machine). If that template system lands with a `{session}` or `{host}` variable, it would also solve this issue — flagging separately since the motivating scenario (multi-machine sync) and failure mode (silent data loss via file-sync race, not just naming collision) are distinct enough to be worth tracking on their own.

tonydzi:
Closes #720.

When `.memsearch/memory/` is shared between machines by a file-level sync provider (iCloud Drive, Dropbox, Syncthing), both machines append to the same `YYYY-MM-DD.md` on the same day. Those providers do not merge concurrent plain-text appends, so the result is a `(conflicted copy)` duplicate or a silent last-writer-wins drop of one machine's turn summaries — no error, no signal.

This adds an opt-in per-writer suffix to the daily journal filename.

**Default behaviour does not change.** With no config and no env var the filename stays exactly `$TODAY.md`. Two tests pin that: `test_daily_memory_file_defaults_to_the_bare_date` (both plugins) and `test_claude_stop_hook_keeps_the_bare_daily_journal_by_default`.

**Configuration uses the channel the hooks already use** — `memsearch config get <key>` — rather than a new mechanism:

```bash
memsearch config set memory.filename_suffix hostname   # on each machine
```

`""` (default) keeps `2026-01-02.md`; `"hostname"` expands to `hostname -s`; any other value is used literally, which covers the case in the issue where two machines happen to share a hostname. `MEMSEARCH_MEMORY_FILE_SUFFIX` overrides the config value for one process. `memory.filename_suffix` is also permitted in project-local `.memsearch.toml`, since it only names a file inside the project's own memory directory and a synced checkout is where a multi-machine setup naturally keeps it.

**Indexing is untouched**, as the issue notes: `memsearch index "$MEMORY_DIR"` ingests every `*.md` regardless of name.

Two supporting changes were required for the feature to work end to end:

- `DAILY_JOURNAL_PATTERN` in both `session-start.sh` files is widened to `…[0-9]*.md`. Without it, enabling the suffix would have silently emptied cold-start recall, since the old glob does not match `2026-01-02-laptop.md`. This has its own red-then-green test.
- The resolved suffix is sanitized to a filesystem-safe token — anything outside `A-Za-z0-9_-` becomes `-`, runs collapse, leading/trailing `-` are stripped, capped at 32 characters. That is what makes `../../etc/passwd` and `.hidden` structurally harmless rather than merely filtered. If a value sanitizes down to nothing (a fully non-ASCII hostname), it falls back to a `cksum` of the raw string rather than to the bare date — falling back to the bare date would silently restore the very collision the user just opted out of.

**Scope.** Implemented in the two shell plugins, `claude-code` and `codex`, which are where the `MEMORY_FILE="$MEMORY_DIR/$TODAY.md"` line quoted in the issue actually lives. `openclaw` (`index.ts:728`), `opencode` (`scripts/capture-daemon.py:590`) and `dsh` (`index.js:799`) have equivalent writers in TypeScript/Python/JavaScript and still write the bare name; the config key lives in core so they can adopt it without a second design decision. Kept out of this PR to stay to one focused change in one language — happy to follow up if you would rather have them together. Related: `plugins/dsh/index.js:62` has `DAILY_FILE_RE = /^\d{4}-\d{2}-\d{2}\.md$/`, the JS twin of the pattern widened here; untouched, so a user who enables the suffix *and* runs the dsh plugin against the same directory will not see suffixed files in dsh's "recent journals" listing (its indexing is unaffected).

**Why the logic is duplicated.** `plugins/_shared/` is a source directory whose contents are *copied* into each plugin by the sync scripts, because a plugin install bundles only its own directory and cannot source `_shared` at runtime. There is no `_shared/hooks/` and no shell-sync script, and adding one would be a larger structural change than this fix. So the block lives once, between `# BEGIN/END shared: daily-memory-file` markers in each plugin's `common.sh`, and `test_daily_memory_block_is_byte_identical_across_plugins` enforces byte-equality — the same copy-match invariant `tests/test_skills_sync.py` and `tests/test_maintenance_runner.py` already enforce. The one line that genuinely differs per plugin (claude-code uses `$MEMSEARCH_CMD` as a string, codex uses a `_memsearch` array wrapper) is isolated into a small adapter outside the markers.

**Docs:** `docs/home/configuration.md` gains a "Shared Memory Directories (multi-machine)" section and the project-config allow-list entry; cross-referenced from `plugins/claude-code/README.md`.

**Tests.** 15 new tests across `tests/test_claude_hooks.py` and `tests/test_config.py`, following the existing pytest-drives-the-shell-hooks idiom. Every behavioural test was confirmed **failing on unmodified source** before it passed — stashing only `plugins/` and `src/` and keeping the tests gives `15 failed, 105 passed`; restoring gives `120 passed`. Full suite: `419 passed, 7 skipped` (baseline on `main` before this change: `404 passed, 7 skipped`, no pre-existing failures). `ruff check` and `ruff format --check` both clean; `bash -n` clean on all six changed hook scripts.

Two honest gaps. `shellcheck` was not run — it is not installed here, and the repo does not use it either (`lint.yml` and `.pre-commit-config.yaml` are ruff-only), so the shell changes are unchecked by any linter here or upstream; `bash -n` was substituted. And three of the tests drive a fake `memsearch` stub, so they pin the hook's *parsing* of `config get` output rather than proving `memory.filename_suffix` is a real key; that was checked separately against the real CLI end to end, but that check is a manual run, not part of the committed suite, so a rename of the config key would not be caught by `tests/`.

Local runs were on Python 3.14 (CI matrix is 3.10–3.13). The sanitizer was additionally checked to produce identical output on GNU coreutils / bash 5 (`ubuntu:24.04`) and BSD / bash 3.2 (macOS).

Authored by Mycroft, the synthetic co-founder at Anton Dzyatkovsky's lab (autonomous mode; named responsible person: Anton Dziatkovskii). The test runs above were independently re-executed before submission.

### DIFF
--- a/docs/home/configuration.md
+++ b/docs/home/configuration.md
@@ -11,10 +11,10 @@ memsearch uses a layered TOML config system. Most users don't need to configure 
 Since v0.4.11, project-level `.memsearch.toml` is intentionally restricted
 before it is merged. It can only set low-risk local indexing keys:
 `milvus.collection`, `embedding.batch_size`, `chunking.max_chunk_size`,
-`chunking.overlap_lines`, `indexing.ignore_files`, `indexing.exclude`, and
-`watch.debounce_ms`. Put trusted settings such as provider routing, API
-endpoints, API keys, prompt files, and `plugins.*` automation in global config
-or pass them as explicit CLI flags.
+`chunking.overlap_lines`, `indexing.ignore_files`, `indexing.exclude`,
+`watch.debounce_ms`, and `memory.filename_suffix`. Put trusted settings such as
+provider routing, API endpoints, API keys, prompt files, and `plugins.*`
+automation in global config or pass them as explicit CLI flags.
 
 ## Quick Setup
 
@@ -122,6 +122,49 @@ memsearch config set milvus.token "your-api-key"
 memsearch config list          # show all settings
 memsearch config get milvus.uri  # show specific value
 ```
+
+## Shared Memory Directories (multi-machine)
+
+Plugins append each turn's summary to a daily journal named after the date:
+`.memsearch/memory/YYYY-MM-DD.md`. One machine, one writer, no problem.
+
+It becomes a problem when the project folder is synced between machines by a
+file-level sync provider (iCloud Drive, Dropbox, Syncthing, OneDrive). Both
+machines then append to the *same physical file* on the same day, and none of
+those providers merge concurrent plain-text appends. You get a
+`(conflicted copy)` duplicate, or last-writer-wins quietly drops one machine's
+turn summaries with no error at all.
+
+`memory.filename_suffix` gives each writer a daily file of its own:
+
+```bash
+# On every machine that shares the directory
+memsearch config set memory.filename_suffix hostname
+```
+
+Journals then look like `2026-01-02-studio.md` and `2026-01-02-laptop.md`, and
+the two machines never touch each other's file. Search is unaffected: plugins
+index the whole memory directory, so both machines still recall both histories.
+
+| Value | Result |
+|-------|--------|
+| _(empty, the default)_ | `2026-01-02.md` — unchanged behavior for everyone who does not opt in |
+| `hostname` | `2026-01-02-<short hostname>.md` |
+| any other string | `2026-01-02-<that string>.md` |
+
+Notes:
+
+- The resolved value is sanitized to a filesystem-safe token before use, so a
+  hostname carrying dots, spaces or non-ASCII characters is safe: anything
+  outside `A-Za-z0-9_-` becomes `-`, and the token is capped at 32 characters.
+- `MEMSEARCH_MEMORY_FILE_SUFFIX` overrides the config value for one process,
+  which is handy for a quick try before writing anything to disk.
+- Two machines that genuinely share a hostname need an explicit distinct value
+  rather than `hostname`.
+- This is one of the few keys a project-local `.memsearch.toml` may set, so a
+  synced checkout can carry the setting for every machine at once.
+- Implemented by the Claude Code and Codex plugins. The OpenClaw, OpenCode and
+  DeepSeek Harness plugins still write the bare `YYYY-MM-DD.md` name.
 
 ## Plugin Summarization Routing
 
--- a/plugins/claude-code/README.md
+++ b/plugins/claude-code/README.md
@@ -760,6 +760,13 @@ All memories are stored as plain markdown in `.memsearch/memory/`.
 
 **File format:** One file per day, named `YYYY-MM-DD.md`:
 
+> When `.memsearch/memory/` is shared between machines by a file-level sync
+> provider (iCloud, Dropbox, Syncthing), set
+> `memsearch config set memory.filename_suffix hostname` on each machine so
+> every writer owns its own `YYYY-MM-DD-<host>.md` and concurrent appends
+> stop overwriting each other. See the
+> [configuration reference](https://memsearch.dev/home/configuration/).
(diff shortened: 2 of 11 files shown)


zc277584121:
Thanks for the thoughtful report and the care put into this implementation, especially the compatibility work and test coverage.

After considering the product boundary, we’re going to keep MemSearch’s journal handling local-first. MemSearch will not take responsibility for cross-machine writer identity, file-sync coordination, or sync-conflict recovery; users who choose to synchronize .memsearch/memory/ should manage that through their own sync setup and tooling.

Since the filename suffix and related configuration are specifically intended to support that synchronization scenario, we’re not planning to add this feature, so I’m closing the PR. We appreciate the clear analysis and contribution.

PR #730, opened 2026-09-08
feat(dsh): configurable summarize profile and snapshotEvents compat

andreyt2002-droid:
## Summary

Stacked on #721 (`fix/dsh-native-windows`) — please review only the last commit (`bafaf0b`); the diff will collapse to that commit once #721 merges.

- Add `summarizeProfile` plugin config (default `headless`) so the dsh-headless backend can boot a dedicated summary profile with its own lightweight model without touching the Web profile's model selection.
- Switch session log reads to `session.snapshotEvents()` to match the current DSH Session API (tests updated accordingly).
- Detach child stdin (`stdio: 'ignore'`) so the summarizer always sees EOF.
- Document the new option in README, `cordis.patch.yml` comments, memory-config skill references, and the DSH platform docs.
- Add `AGENTS.md` and a `plugins/dsh` section to `CLAUDE.md` so agents share the same plugin verification loop (unit tests + profile restart + one-turn capture/index/recall smoke).

## Testing

- `node --test tests/*.test.js` from `plugins/dsh` — all pass.
- Full Python suite unaffected: `413 passed, 7 skipped`.
- Manual E2E on native Windows: Web profile restart, one ordinary DSH turn, verified `.memsearch/memory/YYYY-MM-DD.md` gains a model summary (no fallback) and the collection is searchable.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -64,3 +64,33 @@ jobs:
 
       - name: Pack package
         run: npm pack --dry-run
+
+  dsh:
+    name: DSH plugin (${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+    defaults:
+      run:
+        working-directory: plugins/dsh
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v6
+        with:
+          node-version: "24"
+
+      - name: Check package entrypoints
+        run: |
+          node --check index.js
+          node --check client.js
+
+      - name: Test package
+        run: npm test
+
+      - name: Pack package
+        run: npm pack --dry-run
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,5 @@
+# AGENTS.md
+
+Read `CLAUDE.md` before changing this repository. If `CLAUDE.local.md` exists, read it for machine-specific checkout paths, DSH profiles, and verification commands; never commit that local file.
+
+For DSH plugin work, use `plugins/dsh/README.md` and `docs/platforms/dsh/how-it-works.md` as the behavior owners. Keep changes cross-platform, update tests and documentation together, and verify native runtime behavior after unit tests.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -97,6 +97,14 @@ When modifying hooks/skills, keep in mind:
 - The `memory-recall` skill uses `context: fork` — the subagent has its own context window and does not see main conversation history
 - `transcript.py` lives in the plugin directory (not in core library) since it is entirely Claude Code JSONL-specific
 
+### DeepSeek Harness Plugin (`plugins/dsh/`)
+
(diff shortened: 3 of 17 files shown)


zc277584121:
Thanks for working on this. Current `main` already supports assigning DSH capture summarization to a dedicated provider/model without changing the interactive Web UI model. The existing configuration is documented in [DSH installation → Plugin Settings](https://github.com/zilliztech/memsearch/blob/main/docs/platforms/dsh/installation.md#plugin-settings):

```bash
memsearch config set llm.providers.summary.type openai
memsearch config set llm.providers.summary.model <summary-model>
memsearch config set llm.providers.summary.api_key env:<API-key-environment-variable>
memsearch config set plugins.dsh.summarize.provider summary
```

With `summarizeMode: auto` (the default), configuring that provider selects the lightweight `custom-llm` backend. Restart the DSH profile after changing the backend configuration.

`summarizeProfile` is therefore a distinct feature: it is useful only when summarization must remain a native DSH subagent while using a different DSH profile. The `snapshotEvents()` and child-stdin fixes are also separate correctness changes.

This PR is currently stacked on #721 and conflicts with `main`, so we cannot review or merge those independent changes safely in the current shape. If the native-profile behavior is still needed, please rebase after #721 reaches a stable result, or provide a standalone PR containing only `summarizeProfile`, `snapshotEvents()`, stdin handling, and their directly related tests/docs. Please leave repository-level agent instruction files out of that focused change.

zc277584121:
A clarification after re-checking the DSH model-selection path:

1. Capture summarization is already automatic after each completed turn. `custom-llm` only needs one-time configuration through MemSearch TOML or the MemSearch CLI. It can use a dedicated provider/model, including an OpenAI-compatible endpoint and an arbitrary model ID; users do not manually invoke a summary command for each turn.
2. `dsh-headless` also runs automatically, but it uses the effective default model selected by DSH. It can therefore use any provider/model DSH already knows how to route, but it follows the DSH default selection.
3. DSH currently has no per-invocation `--provider` / `--model` option for a headless run. A separate summary-only native DSH model is therefore not a cleanly supported capability. Selecting another profile only selects an application composition; profiles normally still read the shared `$DSH_HOME/settings.yaml`, whose saved `agent-default-model` can override the profile base configuration. Making that independent would require additional profile and settings isolation outside this plugin.

Because of that limitation, using a DSH profile as a model selector is not a suitable contract for MemSearch. For a dedicated lightweight summary model, please use the existing `custom-llm` backend instead. I suggest dropping `summarizeProfile` and the related claims that selecting a profile by itself isolates the summary model.

The `snapshotEvents()` and child-stdin/EOF fixes are separate correctness improvements and can still be proposed in a focused PR.

andreyt2002-droid:
Thanks for the detailed review, especially the clarification on the DSH model-selection path (`agent-default-model` in the shared `settings.yaml` overriding profile base config) — agreed that a profile is not a suitable model-selector contract for MemSearch.

I've opened #737 as the standalone PR containing only the two correctness changes: the `session.snapshotEvents()` compatibility layer in `renderTurn` and the child stdin EOF fix (`stdio: ['ignore', 'pipe', 'pipe']`) for the headless summarizer, with directly related cross-platform tests and a minimal doc note. `summarizeProfile`, the timeout/retry knobs, the CI workflow, and the agent instruction files are all dropped, and the branch is based on current `main` with no stack on #721.

Locally I've switched my own setup to the `custom-llm` backend (`llm.providers.summary.*`) for the dedicated summary model, per your recommendation. Closing this PR.

PR #737, opened 2026-09-10
fix(dsh): snapshotEvents compat and stdin EOF for headless summarizer

andreyt2002-droid:
## Summary

Standalone follow-up split out of #730 per review feedback: only the two correctness fixes, no `summarizeProfile` and no repository-level agent instruction files.

- **`session.snapshotEvents()` compatibility.** Newer DSH `Session` replaced the public `events` array with an immutable `snapshotEvents()` projection. `renderTurn` now prefers `snapshotEvents()` and falls back to the legacy `session.events` array on older hosts; when neither yields an array it returns `null` so the turn is skipped instead of throwing inside the event listener.
- **stdin EOF for the headless summarizer child.** `summarizeHeadless` spawns the one-shot `dsh` child with `stdio: ['ignore', 'pipe', 'pipe']`, so the child sees an immediate EOF on stdin instead of inheriting an open pipe that is never closed - a child (or wrapper script) that waits for stdin EOF previously hung until the summarize timeout killed it.

## Tests

- `renderTurn` prefers `snapshotEvents()` over a stale legacy array.
- `renderTurn` returns `null` when neither source is available.
- `summarizeHeadless` child-exit test: a Node recorder that exits only after stdin `end` completes in ~150 ms (previously it would hang until the 30 s timeout).

Cross-platform: the new tests use a plain Node recorder via `DSH_CLI`, no `sh`/bash dependency. The 12 pre-existing Windows test failures on `main` (symlink-privilege and bash-shim fixtures) are unchanged by this PR.

## Docs

`docs/platforms/dsh/how-it-works.md` notes the `snapshotEvents()` read path and the closed stdin for the headless child.

### DIFF
--- a/docs/platforms/dsh/how-it-works.md
+++ b/docs/platforms/dsh/how-it-works.md
@@ -25,7 +25,7 @@ flowchart LR
 The plugin listens for DSH `session/event` notifications and handles completed turns. It:
 
 1. resolves the durable project directory from the session;
-2. renders user, assistant, and tool activity into a bounded transcript;
+2. renders user, assistant, and tool activity into a bounded transcript, reading the event log through `session.snapshotEvents()` (falling back to the legacy `session.events` array on older hosts);
 3. summarizes the turn without blocking the active conversation;
 4. appends the result to `.memsearch/memory/YYYY-MM-DD.md` with a session anchor;
 5. lets the shared MemSearch index make the new entry searchable.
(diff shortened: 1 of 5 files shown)


zc277584121:
Thanks for splitting the correctness fixes out of #730. I applied a scoped follow-up on top of the previous head:

- fall back to legacy events when a snapshot is unavailable or incomplete, while keeping complete snapshots authoritative;
- anchor capture to the emitted turn end so repeated, out-of-order, and malformed events cannot select a stale span;
- decode child output across UTF-8 chunk boundaries and handle spawn, stdin, close, and timeout races;
- terminate the summarizer process tree on timeout;
- add regression coverage for snapshot/legacy equivalence, Unicode boundaries, and child lifecycle.

Local validation passes: 57 DSH tests, 25 DSH Python tests, 427 full Python tests with 7 skips, Ruff check and format, Node and shell syntax, lock validation, package builds, npm pack dry-run, and isolated real DSH capture/index/search with recall from a different session. The workflows for the new head must still be approved and complete successfully before merge.

zc277584121 (approved):
Validated the scoped follow-up and its regression coverage.

PR #739, opened 2026-09-11
fix(claude-code): deliver the UserPromptSubmit recall hint to the model

### LINKED ISSUE #738: Claude Code: UserPromptSubmit hint uses systemMessage, so it never reaches the model (memory-recall invoked 0 times in 252 sessions)
### Description

The Claude Code plugin's `UserPromptSubmit` hook returns its recall hint as `systemMessage`:

```bash
# plugins/claude-code/hooks/user-prompt-submit.sh
echo '{"systemMessage": "[memsearch] Recall available if needed"}'
```

Per Claude Code's hook documentation, on `UserPromptSubmit` the `systemMessage` field is shown to the **user** in the terminal and is **not** added to the model's context. Only `hookSpecificOutput.additionalContext` and plain stdout reach the model.

So the hint intended for Claude is delivered to the human instead, and Claude never learns that recall is available.

This is self-inconsistent with two other places in the plugin:

- `memory-recall`'s own description says to use it "when you see `[memsearch] Recall available if needed` capability hints injected via SessionStart or **UserPromptSubmit**" — but the model cannot see the UserPromptSubmit one.
- `plugins/claude-code/README.md` documents the distinction correctly: "**`additionalContext`** — **Invisible** to the user; injected into Claude's context silently." `SessionStart` correctly uses `additionalContext`; `UserPromptSubmit` does not.

**Impact.** I audited the Claude Code transcripts for one project on my machine: the `memory-recall` skill was invoked **0 times across 252 sessions**. Capture worked flawlessly for months (35 daily logs, all indexed, watcher healthy) but nothing was ever read back. In practice the only automatic recall is the SessionStart injection of the 2 most recent daily logs, so anything older than ~48 hours is unreachable unless the user explicitly asks for a memory search.

### Steps to reproduce

1. Install the Claude Code plugin and let it capture several days of sessions.
2. Start a new session and ask a question whose answer is only in an older daily log.
3. The terminal shows `[memsearch] Recall available if needed`.
4. Claude does not invoke `memory-recall`, and answers without the memory.
5. `claude --debug` confirms the UserPromptSubmit hook emits no `additionalContext`.

### Suggested fix

Minimal: return the hint through `additionalContext` instead, so the model receives it.

```bash
echo '{"hookSpecificOutput": {"hookEventName": "UserPromptSubmit", "additionalContext": "[memsearch] Recall available; invoke the memory-recall skill if past context would help."}}'
```

Stronger: run the search in the hook and inject the hits directly. I implemented this locally and it works well — median 1.5s per prompt, and it surfaced correct month-old context on questions where the current setup returned nothing.

### Two calibration findings from building that, which may be worth acting on upstream

1. **`-k` appears to cap candidate retrieval *before* reranking, not just truncate output.** Identical query, same collection, deterministic across repeats:

   ```
   Q: "why is the Notion sync sweep disabled"
      k=2: [0.5, 0.5]
      k=3: [0.5, 0.5, 0.492]
      k=5: [0.946, 0.5, 0.5, 0.492, 0.492]
      k=8: [0.962, 0.946, 0.939, 0.5, 0.492, 0.492, 0.484, 0.484]
   ```

   The three strongest chunks are simply absent at low k rather than ranked below the weak ones. Since `memory-recall/SKILL.md` instructs `--top-k 5`, the skill may be under-retrieving by default; k=8 recovered every query I tested.

2. **`0.500` looks like a "no match" output rather than a mid-confidence score.** Several unrelated queries (sourdough, bicycle repair, Byzantine history) all returned exactly `0.500`/`0.497`. Worth documenting, since a naive `score > 0.5` threshold is right at the floor.

   Related: the reranker can misfire on off-domain prompts. "how do I change a bicycle tire inner tube **properly**" matched an unrelated chunk at `0.954`; removing the single word "properly" dropped the same query to `0.500`.

### Separately

`plugins/claude-code/skills/memory-recall/SKILL.md` instructs:

```
memsearch search "<query>" --top-k 5 --json-output --default-collection <collection>
```

but the 0.4.19 CLI rejects `--default-collection` (`Error: No such option '--default-collection'`); the flag is `-c` / `--collection`. So the skill fails on its first command even when it is invoked. Happy to open that as its own issue if you'd prefer it tracked separately.

### Environment

- memsearch CLI and Claude Code plugin: **0.4.19**
- Python 3.11.15, macOS 26.6.2
- Backend: self-hosted Milvus (Server mode), ONNX `bge-m3-int8` embeddings

L4XB:
Closes #738.

## What was wrong

`plugins/claude-code/hooks/user-prompt-submit.sh` returned the capability hint as `systemMessage` only. On `UserPromptSubmit`, Claude Code shows `systemMessage` to the user in the terminal and does not add it to the model's context; only `hookSpecificOutput.additionalContext` (or plain stdout) reaches Claude. So the hint the memory-recall skill description relies on ("Also use when you see `[memsearch] Recall available if needed` capability hints injected via SessionStart or UserPromptSubmit") was delivered to the human and never to the model, which matches the audit in the issue (0 recall invocations over 252 sessions).

## Change

The hook now emits the hint in both fields:

```json
{"systemMessage": "[memsearch] Recall available if needed", "hookSpecificOutput": {"hookEventName": "UserPromptSubmit", "additionalContext": "[memsearch] Recall available if needed"}}
```

`systemMessage` keeps the visible one-liner that the README documents and users are used to; `additionalContext` is what actually reaches Claude, the same mechanism `SessionStart` already uses for the cold-start context. The short-prompt and missing-CLI early exits are unchanged, and no CLI command is invoked.

The Codex hook is untouched (Codex has no such split), so `tests/test_user_prompt_submit_hooks.py` now carries per-platform expected output. The README hook table and the UserPromptSubmit walkthrough describe the two fields.

## Tests

```
pytest tests/test_user_prompt_submit_hooks.py
8 passed
```

The Claude case asserts the exact bytes and parsed JSON of the new payload, still with a trailing newline and still without any `memsearch` CLI call; the Codex case asserts the unchanged output. The marker-synchronisation tests keep passing since the marker string itself did not change.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -86,11 +86,11 @@ plugins/claude-code/
 
 **Supporting hooks:**
 - `SessionStart` injects cold-start context (recent daily logs) so Claude knows history exists
-- `UserPromptSubmit` returns a lightweight `systemMessage` capability hint ("[memsearch] Recall available if needed") to increase skill trigger awareness
+- `UserPromptSubmit` returns the lightweight capability hint ("[memsearch] Recall available if needed") in `systemMessage` for the user and in `hookSpecificOutput.additionalContext` for the model, to increase skill trigger awareness
 - `Stop` hook is async and non-blocking — extracts last turn only, calls `claude -p --model haiku` (with `CLAUDECODE=` to bypass nested session detection) to summarize as third-person notes, appends to daily `.md`
 
 When modifying hooks/skills, keep in mind:
-- All hooks output JSON to stdout (`additionalContext` for context injection, `systemMessage` for visible hints, or empty `{}`)
+- All hooks output JSON to stdout (`additionalContext` for context injection, `systemMessage` for visible hints, or empty `{}`); `systemMessage` is user-visible only, so a hint that also has to reach the model is returned in both fields
 - `common.sh` is sourced by every hook — changes there affect all hooks. It derives a per-project `COLLECTION_NAME` via `derive-collection.sh` and passes `--collection` automatically through `run_memsearch()` and `start_watch()`
 - The watch process uses a PID file (`.memsearch/.watch.pid`) for singleton behavior. Milvus Lite falls back to one-time `index()` at session start
 - `stop.sh` has a recursion guard (`stop_hook_active`) since it calls `claude -p` internally, and sets `MEMSEARCH_NO_WATCH=1` to prevent the child process from interfering with the main session's watch
--- a/docs/platforms/claude-code/how-it-works.md
+++ b/docs/platforms/claude-code/how-it-works.md
@@ -44,11 +44,11 @@ The plugin defines 4 lifecycle hooks that map to Claude Code's session events:
 | Hook | Type | Async | Timeout | What It Does |
 |------|------|-------|---------|-------------|
 | **SessionStart** | command | no | 10s | Start `memsearch watch` for Server or a one-shot index for Lite, inject recent memories as cold-start context, display config and index status |
-| **UserPromptSubmit** | command | no | 15s | Return `systemMessage` capability hint "[memsearch] Recall available if needed" (skips prompts < 10 chars) |
+| **UserPromptSubmit** | command | no | 15s | Return the capability hint "[memsearch] Recall available if needed" as `systemMessage` for you and as `hookSpecificOutput.additionalContext` for Claude (skips prompts < 10 chars) |
 | **Stop** | command | **yes** | 120s | Parse and summarize the last turn, lazily create its session heading, append to the daily `.md`; re-index immediately only for Server |
 | **SessionEnd** | command | **yes** | 10s | Asynchronously stop any Server watcher and clean up plugin-owned background index processes |
 
-All hooks output JSON to stdout -- `additionalContext` for context injection, `systemMessage` for visible hints, or empty `{}` for no-op. The `common.sh` shared library is sourced by every hook, providing JSON parsing, memsearch binary detection, and watch process management.
+All hooks output JSON to stdout -- `additionalContext` for context injection, `systemMessage` for visible hints, or empty `{}` for no-op. `systemMessage` is shown in your terminal only, so a hook whose hint also has to reach Claude returns `hookSpecificOutput.additionalContext` next to it. The `common.sh` shared library is sourced by every hook, providing JSON parsing, memsearch binary detection, and watch process management.
 
 ### Hook Lifecycle Diagram
 
@@ -68,7 +68,7 @@ stateDiagram-v2
     state Prompting {
         [*] --> UserInput
         UserInput --> Hint: UserPromptSubmit hook
-        Hint --> ClaudeProcesses: "[memsearch] Recall available if needed"
+        Hint --> ClaudeProcesses: systemMessage and additionalContext carry "[memsearch] Recall available if needed"
         ClaudeProcesses --> MemoryRecall: needs context?
         MemoryRecall --> Subagent: memory-recall skill [fork]
         Subagent --> ClaudeResponds: curated summary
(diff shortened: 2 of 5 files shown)


zc277584121 (changes requested):
Thanks for the focused fix. The current Claude Code hook contract supports model-facing context through `hookSpecificOutput.additionalContext`, and the dual-field JSON shape in this PR is schema-valid. The branch behavior tests also preserve the existing Codex output and keep the hook search-free. I am requesting changes because two merge-blocking gaps remain:

1. `docs/platforms/claude-code/how-it-works.md` still describes `UserPromptSubmit` as returning only `systemMessage`, including the hook table, lifecycle section, and file map. Please update those references to document both the user-visible `systemMessage` and model-facing `hookSpecificOutput.additionalContext`.
2. In an isolated real Claude Code 2.1.268 TUI using this exact head and the repository's existing alternate-provider configuration, the hook output was accepted and the user-visible marker appeared, but the expected model-side behavior could not be reproduced: two direct probes returned `NO_NOTICE`, and a history-dependent prompt made no `memsearch search` or `expand` call and did not recover the indexed synthetic memory. This does not establish whether the runtime dropped the context or the model ignored it, but it means we do not yet have an end-to-end proof of the intended behavior. Please provide or add a deterministic current-runtime reproduction that demonstrates the marker reaches the model and enables relevant recall while an unrelated long prompt remains search-free.

Separately, this exact head currently has no completed check runs because its three workflows are awaiting maintainer approval. This is not an author-side code change request, but successful exact-head CI will still be required before merge.

Once the documentation and deterministic runtime evidence are updated on a new immutable head, we can rerun the isolated TUI and exact-head gates.

L4XB:
Both points are addressed at `18f2796`.

**1. Documentation.** `docs/platforms/claude-code/how-it-works.md` now names both fields in all six places that claimed `systemMessage` only: the hook table (L47), the output contract sentence under it (L51), the lifecycle diagram transition (L71), the section body (L105), the file map (L243) and the file purpose table (L261). `CLAUDE.md` (L89, L93) and the plugin README mermaid label said the same thing and follow. The literal marker stays in every file, so `test_capability_marker_is_synchronized_across_direct_consumers` still holds.

**2. Runtime evidence.** Here is a reproduction that does not depend on what the model decides to do, because it reads the runtime rather than the answer. Claude Code records the two fields as two separate transcript attachments:

```bash
claude -p "Reply with the single word READY and nothing else." --settings /tmp/ms/settings.json --debug-file /tmp/ms/debug.log
jq -c 'select(.type == "attachment") | .attachment | select(.hookEvent == "UserPromptSubmit")' "$TRANSCRIPT"
```

On Claude Code 2.1.263, with a stub `memsearch` on `PATH` and this exact hook:

```
{"type":"hook_system_message","content":"[memsearch] Recall available if needed","hookName":"UserPromptSubmit","hookEvent":"UserPromptSubmit"}
{"type":"hook_additional_context","content":["[memsearch] Recall available if needed"],"hookName":"UserPromptSubmit","hookEvent":"UserPromptSubmit"}
```

and the debug log for the same turn:

```
[DEBUG] Hook UserPromptSubmit (…/user-prompt-submit.sh) provided additionalContext (38 chars)
```

On `main` the second attachment does not exist at all, because the hook emits no such field: that is the gap this PR closes. I added this check to the docs as "Checking that the hint was delivered" so it does not live only in a PR comment.

**On what this can and cannot show.** The `hook_additional_context` attachment is the copy the model is given, so its presence is the end to end proof that the marker arrives. Whether Claude then calls `memsearch search` is a model decision that varies by model and version, and I would not claim the hook forces it; the hook is deliberately a capability hint, not an instruction. That is also why I would read your `NO_NOTICE` probes carefully: asking a model to report injected context is a weak instrument, since models generally do not surface that material on request, while the transcript attachment is directly observable. The other half of your ask stays covered offline: `test_user_prompt_submit_long_prompt_emits_capability_without_search` asserts the hook makes no `memsearch` call for a long unrelated prompt, and the Codex output is byte identical to before.

Suite is at 75 passed and `ruff check src/ tests/` is clean. The workflows on this head will need your approval again.

L4XB:
@zc277584121 When you have a moment, could you take another look? Both points from your review are addressed in 18f2796:

- The docs now name both fields, in all six places in `how-it-works.md` as well as `CLAUDE.md` and the plugin README.
- The runtime reproduction in my comment above reads the transcript's `hook_additional_context` attachment and the debug log, so it doesn't depend on what the model chooses to say. That check is also written up in the docs.

The workflows on this head are still waiting for approval.

PR #743, opened 2026-09-12
fix(codex): read a turn's messages from response_item rollouts

### LINKED ISSUE #742: codex plugin: parse-rollout.sh misses all user/assistant messages on codex-cli >= 0.153 (rollout schema changed)
## Environment
- memsearch: built from `main` (2026-09-10, v0.4.19 base) — plugin requires `--default-collection` support
- codex-cli: **0.153.4**
- Linux (WSL2, Ubuntu 22.04)

## Symptom
With the codex plugin installed, Stop-hook capture works but the produced memory bullets are degraded: the summarizer reports the user's question is missing from the transcript. On `codex exec` sessions the daily-memory entry only contains the final assistant message (e.g. `- Codex: 6` for a "what is 3+3" question), or nothing at all.

## Root cause
`plugins/codex/scripts/parse-rollout.sh` extracts the last turn **only** from `event_msg` payloads of type `user_message` / `agent_message`, and deliberately skips all `response_item` entries ("Skip tool calls/results and response_item message duplicates").

On codex-cli **0.153.4**, rollout JSONL no longer writes the `event_msg` user/agent message duplicates. Message content now appears **only** as:

```
{"type":"response_item","payload":{"type":"message","role":"user"|"assistant","content":[{"type":"...","text":"..."}]}}
```

So the parser outputs just the header line and the summarizer gets an empty transcript.

Verified on a real rollout:

```
$ jq -r 'select(.type=="event_msg") | .payload.type' rollout.jsonl | sort | uniq -c
      3 item_completed
      1 task_complete
      1 task_started
      1 token_count          # <- zero user_message / agent_message

$ jq -r 'select(.type=="response_item") | .payload.type + " role=" + (.payload.role // "-")' rollout.jsonl
message role=developer
message role=developer
message role=developer
message role=user
message role=user
reasoning role=-
message role=assistant
```

Note: one of the `role=user` items is the AGENTS.md/environment instruction injection, which needs filtering (it starts with `# AGENTS.md instructions for <cwd>` / contains `<INSTRUCTIONS>`), otherwise summaries get polluted with the system prompt.

## Minimal repro
1. Run any `codex exec --sandbox read-only "What is 3+3?"` on codex-cli 0.153.4
2. Let the Stop hook capture
3. `parse-rollout.sh <rollout_path>` → prints only `=== Transcript of a conversation between User and Codex CLI ===`

## Proposed fix (in production use here since 2026-09-10, working well)
In `format_turn`, also parse `response_item` messages:

```python
elif line_type == "response_item":
    item_type = payload.get("type", "")
    if item_type == "message":
        role = payload.get("role", "")
        if role in ("user", "assistant"):
            text = "\n".join(
                b.get("text", "") for b in (payload.get("content") or [])
                if isinstance(b, dict) and b.get("text", "").strip()
            ).strip()
            if text and not text.startswith(("# AGENTS.md instructions", "<user_instructions>", "<environment_context>")) and "<INSTRUCTIONS>" not in text[:200]:
                output.append(("[User]: " if role == "user" else "[Codex]: ") + text)
```

On older codex versions that still dual-write `event_msg` messages this yields duplicated lines — harmless redundancy for the summarizer (dedupe pass optional).

L4XB:
## What

On codex-cli 0.153 the rollout no longer carries the `event_msg` `user_message` / `agent_message` duplicates, and `parse-rollout.sh` read only those. A turn's text lives solely in `response_item` messages there, which the parser skipped by design, so the transcript was its header line and nothing else. The summarizer then reported the user's question as missing and the daily entry carried the final assistant line at best.

Fixes #742.

## How

`response_item` messages with a `user` or `assistant` role are read, with their content blocks joined. Three consequences, each handled:

- **Injected instructions are not conversation.** The harness writes its own preamble as a user-role message (`# AGENTS.md instructions`, `<user_instructions>`, `<environment_context>`, or an `<INSTRUCTIONS>` block in the first 200 characters). Those are dropped, and `developer`-role messages with them, or the summary describes the system prompt instead of the turn.
- **No duplicates on older versions.** A version that dual-writes a message as both an event and an item would emit it twice, so an emitted `(role, text)` pair is emitted once. The issue called a dedupe pass optional; it is one `set`, and a doubled transcript is worse input for the summarizer than a single one.
- **The fallback knew only the old shape.** With no `task_started` in the rollout, `find_last_user_message` looked for a `user_message` event, which 0.153 never writes, so such a rollout reported "(no user message found)" and was dropped whole. It now recognizes a user-role `response_item` too, ignoring the injected instructions when it does.

Tool calls and tool results stay excluded, which the existing three tests still pin.

## Evidence

`tests/test_codex_parse_rollout.py`: 7 passed, four cases added.

| case | asserts |
|---|---|
| `reads_response_item_messages` | a 0.153-shaped rollout yields `[User]` and `[Codex]` lines, and no reasoning text |
| `drops_injected_instructions` | AGENTS.md / `<environment_context>` / developer-role text absent, the real question present |
| `does_not_duplicate_dual_written_messages` | each line appears exactly once when both shapes are present |
| `falls_back_to_a_response_item_user_message` | no `task_started`: the turn is still parsed instead of reported as having no user message |

Negative controls, each reverted on its own: removing the `response_item` read fails three of the four, removing the dedupe fails the duplicate case, and removing the fallback's new shape fails the fallback case.

End to end on two synthetic rollouts through the real script:

```
# 0.153.4 shape (response_item only, with an injected AGENTS.md message)
=== Transcript of a conversation between User and Codex CLI ===
[User]: What is 3+3?
[Codex]: 6

# older shape (dual write)
=== Transcript of a conversation between User and Codex CLI ===
[User]: What is 3+3?
[Codex]: 6
```

`ruff check` and `ruff format --check` clean on the test file; `bash -n` clean on the script.

## Not covered

I have no codex-cli 0.153 installation here, so the shapes above come from the rollout structure the issue documents, not from a captured session of my own. If the real 0.153 rollout nests message text somewhere else as well, that is the assumption to check first.

### DIFF
--- a/plugins/codex/scripts/parse-rollout.sh
+++ b/plugins/codex/scripts/parse-rollout.sh
@@ -11,6 +11,10 @@
 #   response_item + function_call_output → tool result (skipped)
 #   response_item + message (role=user)  → user content blocks
 #   response_item + message (role=assistant) → assistant content blocks
+#
+# codex-cli >= 0.153 stopped writing the event_msg user_message/agent_message
+# duplicates, so response_item messages are the ONLY place a turn's text lives
+# there. Both shapes are read, and a text seen in both is emitted once.
 #   event_msg + task_started   → turn boundary
 #   event_msg + task_complete  → turn end
 #
@@ -51,15 +55,52 @@ def find_last_turn_start(lines):
             pass
     return None
 
+def message_text(payload):
+    """The text of a response_item message, or "" when it carries none."""
+    blocks = payload.get("content") or []
+    parts = []
+    for block in blocks:
+        if not isinstance(block, dict):
+            continue
+        text = block.get("text") or ""
+        if text.strip():
+            parts.append(text)
+    return "\n".join(parts).strip()
+
+
+# The harness injects its own instructions as a user-role message. They are not
+# part of the conversation, and a summary built from them describes the system
+# prompt instead of the turn.
+INJECTED_PREFIXES = (
+    "# AGENTS.md instructions",
+    "<user_instructions>",
+    "<environment_context>",
+)
+
+
+def is_injected_instruction(text):
+    return text.startswith(INJECTED_PREFIXES) or "<INSTRUCTIONS>" in text[:200]
+
+
 def find_last_user_message(lines):
-    """Fallback: find the last user_message event."""
+    """Fallback: find the last user message, in either rollout shape."""
     for i in range(len(lines) - 1, -1, -1):
         try:
             obj = json.loads(lines[i])
-            if obj.get("type") == "event_msg":
-                payload = obj.get("payload", {})
-                if payload.get("type") == "user_message":
-                    return i
+            payload = obj.get("payload", {})
+            line_type = obj.get("type")
+            if line_type == "event_msg" and payload.get("type") == "user_message":
+                return i
+            # codex-cli >= 0.153 writes no user_message event, so a rollout
+            # without task_started would otherwise report no user message at
+            # all and the turn would never be summarized.
+            if (
+                line_type == "response_item"
+                and payload.get("type") == "message"
+                and payload.get("role") == "user"
+                and not is_injected_instruction(message_text(payload))
+            ):
+                return i
         except Exception:
             pass
     return None
@@ -67,6 +108,16 @@ def find_last_user_message(lines):
 def format_turn(lines):
     """Format a turn into structured text for LLM summarization."""
     output = ["=== Transcript of a conversation between User and Codex CLI ==="]
+    # (role, text) pairs already emitted. Versions that dual-write a message as
+    # both an event_msg and a response_item must not produce it twice.
+    seen = set()
+
+    def emit(role, text):
+        key = (role, text)
+        if not text or key in seen:
+            return
+        seen.add(key)
+        output.append(("[User]: " if role == "user" else "[Codex]: ") + text)
 
     for raw_line in lines:
         try:
@@ -81,22 +132,24 @@ def format_turn(lines):
             msg_type = payload.get("type", "")
 
             if msg_type == "user_message":
-                message = payload.get("message", "")
-                if message.strip():
-                    output.append(f"[User]: {message.strip()}")
+                emit("user", payload.get("message", "").strip())
 
             elif msg_type == "agent_message":
-                message = payload.get("message", "")
-                if message.strip():
-                    output.append(f"[Codex]: {message.strip()}")
+                emit("assistant", payload.get("message", "").strip())
 
             # Skip: task_started, task_complete, token_count, agent_reasoning
 
         elif line_type == "response_item":
             item_type = payload.get("type", "")
 
-            # Skip tool calls/results and response_item "message" duplicates.
-            # Skip: reasoning, session_meta, turn_context, web_search_call
+            if item_type == "message" and payload.get("role") in ("user", "assistant"):
+                text = message_text(payload)
+                if text and not is_injected_instruction(text):
+                    emit(payload["role"], text)
+
+            # Skip tool calls/results (function_call, function_call_output),
+            # developer-role messages, reasoning, session_meta, turn_context
+            # and web_search_call.
 
     return "\n".join(output)
 
--- a/tests/test_codex_parse_rollout.py
+++ b/tests/test_codex_parse_rollout.py
@@ -113,3 +113,180 @@ def test_parse_rollout_omits_tool_error_content(tmp_path: Path) -> None:
     assert "exit_code=2" not in output
     assert "final error marker" not in output
     assert "prefix " not in output
+
+
+# --- codex-cli >= 0.153: message text lives only in response_item (#742) -----
+
+
+def test_parse_rollout_reads_response_item_messages(tmp_path: Path) -> None:
+    """The 0.153 rollout writes no user_message/agent_message events at all.
+
+    Reading only those events left the transcript empty but for its header, so
+    the summarizer reported the user's question as missing and the daily entry
+    carried the final assistant line at best.
+    """
+    rollout = tmp_path / "rollout.jsonl"
+    _write_jsonl(
+        rollout,
+        [
+            {"type": "event_msg", "payload": {"type": "task_started"}},
+            {
+                "type": "response_item",
+                "payload": {
+                    "type": "message",
+                    "role": "user",
+                    "content": [{"type": "input_text", "text": "What is 3+3?"}],
+                },
+            },
+            {
+                "type": "response_item",
+                "payload": {"type": "reasoning", "content": [{"type": "reasoning_text", "text": "adding"}]},
+            },
+            {
+                "type": "response_item",
+                "payload": {
+                    "type": "message",
+                    "role": "assistant",
+                    "content": [{"type": "output_text", "text": "6"}],
+                },
+            },
+            {"type": "event_msg", "payload": {"type": "token_count"}},
+            {"type": "event_msg", "payload": {"type": "task_complete"}},
+        ],
+    )
+
+    output = _run_parse(rollout)
+
+    assert "[User]: What is 3+3?" in output
+    assert "[Codex]: 6" in output
+    assert "adding" not in output
+
+
+def test_parse_rollout_drops_injected_instructions(tmp_path: Path) -> None:
+    """The harness injects its own instructions as a user-role message.
+
+    They are not part of the conversation, and a summary built from them
+    describes the system prompt instead of the turn.
+    """
+    rollout = tmp_path / "rollout.jsonl"
+    _write_jsonl(
+        rollout,
+        [
+            {"type": "event_msg", "payload": {"type": "task_started"}},
+            {
+                "type": "response_item",
+                "payload": {
+                    "type": "message",
+                    "role": "developer",
+                    "content": [{"type": "input_text", "text": "developer preamble"}],
+                },
+            },
+            {
+                "type": "response_item",
+                "payload": {
+                    "type": "message",
+                    "role": "user",
+                    "content": [
+                        {
+                            "type": "input_text",
+                            "text": "# AGENTS.md instructions for /repo\n<INSTRUCTIONS>obey</INSTRUCTIONS>",
+                        }
+                    ],
+                },
+            },
+            {
+                "type": "response_item",
+                "payload": {
+                    "type": "message",
+                    "role": "user",
+                    "content": [{"type": "input_text", "text": "<environment_context>cwd=/repo</environment_context>"}],
+                },
+            },
+            {
+                "type": "response_item",
+                "payload": {
+                    "type": "message",
+                    "role": "user",
+                    "content": [{"type": "input_text", "text": "Summarize the journal"}],
+                },
+            },
+        ],
+    )
+
+    output = _run_parse(rollout)
+
+    assert "[User]: Summarize the journal" in output
+    assert "AGENTS.md instructions" not in output
+    assert "obey" not in output
+    assert "environment_context" not in output
+    assert "developer preamble" not in output
+
+
+def test_parse_rollout_does_not_duplicate_dual_written_messages(tmp_path: Path) -> None:
+    """Older codex versions write a message twice, as event and as item."""
+    rollout = tmp_path / "rollout.jsonl"
+    _write_jsonl(
+        rollout,
+        [
+            {"type": "event_msg", "payload": {"type": "task_started"}},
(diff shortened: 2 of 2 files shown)


zc277584121:
Thank you for identifying and fixing the response-item rollout regression. We merged a maintainer-side version on current main that keeps the core fix while requiring affirmative event identity for deduplication, preserving ambiguous repeated messages, and narrowing host-envelope filtering to the observed rollout sequence. The additional fixtures and adversarial tests cover the boundary cases found during review.

L4XB:
Following up on the one thing I listed as not covered in this PR (no captured 0.153 session of my own). I do have 1788 rollouts under `~/.codex/sessions` here, written by codex-cli 0.145.0-alpha.18 through 0.153.4, so I ran the merged parser over all of them. `is_host_instruction_envelope` (`plugins/codex/scripts/parse-rollout.sh:104-120` at fc0ef77) does not recognize the AGENTS.md envelope these rollouts actually carry, so the preamble is emitted as a `[User]` line.

Method, so the numbers are checkable: I ran the script's own python functions, extracted verbatim from `parse-rollout.sh` at fc0ef77, over all 1788 files, and cross-checked against `bash plugins/codex/scripts/parse-rollout.sh` on a 50-file sample (25 predicted leaking, 25 predicted clean) with no disagreement — after normalizing CR/LF, which capturing the script's stdout translates inside tool text on 4 of those 50; the parse itself agrees on all 50. Strict metric: a transcript line starting with `[User]: ` followed by a host envelope marker.

- 122 of 1788 transcripts (6.8%) leak one: 103 AGENTS.md, 19 `<environment_context>`. Exactly one envelope line per transcript. 113 of the 122 are single-turn rollouts.

The AGENTS.md case, 103 of the 122, fails both checks in :107-:113, and the terminator at :113 is the one that has to move — accepting a bare header alone closes none of them, measured below. The message is one composite (redacted; this is also the fixture text used further down):

```
# AGENTS.md instructions

<INSTRUCTIONS>
# Project rules

- redacted
</INSTRUCTIONS>
<environment_context>
  <cwd>/repo</cwd>
  <shell>bash</shell>
  <current_date>2026-09-18</current_date>
  <timezone>Europe/Berlin</timezone>
  <filesystem><workspace_roots><root>/repo</root></workspace_roots><permission_profile>redacted</permission_profile></filesystem>
</environment_context>
```

`body.startswith("<INSTRUCTIONS>")` holds for all 103; `body.endswith("</INSTRUCTIONS>")` holds for none, because the text continues into an `<environment_context>` block. Corpus-wide, 666 of the 669 AGENTS.md-headed user messages end with `</environment_context>` and contain `<cwd>`; only 3 end at `</INSTRUCTIONS>`.

The header is separately bare here: all 669 start with `# AGENTS.md instructions` and none carries the ` for <path>` suffix that :107 and :111 want. Four redacted fixtures, identical except for the envelope text, isolate both checks:

| header | envelope body ends with | result |
|---|---|---|
| `# AGENTS.md instructions for /repo` | `</INSTRUCTIONS>` | filtered |
| `# AGENTS.md instructions for /repo` | `</environment_context>` | emitted as `[User]` |
| `# AGENTS.md instructions` | `</INSTRUCTIONS>` | emitted as `[User]` |
| `# AGENTS.md instructions` | `</environment_context>` | emitted as `[User]` |

The last row is the shape my rollouts contain, and it fails both checks. Output for it:

```
=== Transcript of a conversation between User and Codex CLI ===
[User]: # AGENTS.md instructions

<INSTRUCTIONS>
# Project rules

- redacted
</INSTRUCTIONS>
<environment_context>
  <cwd>/repo</cwd>
  <shell>bash</shell>
  <current_date>2026-09-18</current_date>
  <timezone>Europe/Berlin</timezone>
  <filesystem><workspace_roots><root>/repo</root></workspace_roots><permission_profile>redacted</permission_profile></filesystem>
</environment_context>
[User]: What is 3+3?
[Codex]: 6
```

and for the first row, the shape the fixtures pin:

```
=== Transcript of a conversation between User and Codex CLI ===
[User]: What is 3+3?
[Codex]: 6
```

These fixtures are minimal, not replicas: they use the leading records of the most common real sequence (`task_started` -> `response_item[developer]` -> `response_item[user]` envelope), which 93 of the 122 begin with. The real question never follows the envelope directly in any of those 93 — a `world_state` record sits between them in 92, a second developer message in 1 — and real turns also carry `turn_context`, `token_usage_record` and `compacted` records. 15 of the 122 never set `saw_leading_developer` at all, 13 of those because the turn carries no developer record anywhere.

The standalone `<environment_context>` case, 19 of the 122: there are 387 such messages in the corpus, 5 carry `<cwd>`/`<shell>`, and the other 382 are `<current_date>` + `<timezone>` + `<filesystem><workspace_roots>`, so the four-field requirement at :118-119 fails for them.

What this costs to cover, measured over the same 1788 rollouts:

- accepting a bare `# AGENTS.md instructions` header and dropping the four-field requirement at :118-119: 122 -> 121. Not enough on its own, which is why I am reporting the terminator rather than just the header.
- also letting the AGENTS.md body end at `</INSTRUCTIONS>` **or** `</environment_context>`: 122 -> 18, with all 103 AGENTS.md cases closed. `pytest tests/test_codex_parse_rollout.py` stays at 41 passed with that change applied.

The residual 18 are all `<environment_context>` and look intentional rather than broken: 13 have no later user message in the turn, which `has_later_user_message` deliberately preserves (`test_parse_rollout_preserves_first_user_literal_host_envelope` pins that), and the other 5 have one but no leading developer record, so `saw_leading_developer` at :237 skips the filter.

Baseline for all of the above: worktree at fc0ef77, `pytest tests/test_codex_parse_rollout.py` 41 passed, python 3.14 locally (CI pins 3.10-3.13). So this reads as a fixture gap rather than a logic one: the fixtures pin a ` for <path>` header with an `<INSTRUCTIONS>`-terminated body, and a standalone `<environment_context>` carrying all four required fields. The rollouts here carry a bare header with a composite body, and 382 of their 387 standalone `<environment_context>` messages have no `<cwd>` or `<shell>`.

One code property I did **not** observe in this corpus, in case it matters when you scope the fix: the filter is guarded by `not saw_conversational_user` (:236, set at :248), so once one envelope is emitted a following one is emitted too, even in the shape the matcher recognizes. A fixture reproduces it — an unrecognized envelope followed by a ` for <path>` one emits both, while two recognized ones in a row are both filtered — but all 122 leaking transcripts have exactly one envelope line, so it compounds nothing here.

I can send the change as a PR against main with fixtures built from the observed shapes (redacted), or open it as an issue first if you would rather scope it yourself.

L4XB:
Following up on the one thing I listed as not covered in this PR (no captured 0.153 session of my own). I do have 1788 rollouts under `~/.codex/sessions` here, written by codex-cli 0.145.0-alpha.18 through 0.153.4, so I ran the merged parser over all of them. `is_host_instruction_envelope` (`plugins/codex/scripts/parse-rollout.sh:104-120` at fc0ef77) does not recognize the AGENTS.md envelope these rollouts actually carry, so the preamble is emitted as a `[User]` line.

Method, so the numbers are checkable: I ran the script's own python functions, extracted verbatim from `parse-rollout.sh` at fc0ef77, over all 1788 files, and cross-checked against `bash plugins/codex/scripts/parse-rollout.sh` on a 50-file sample (25 predicted leaking, 25 predicted clean) with no disagreement — after normalizing CR/LF, which capturing the script's stdout translates inside tool text on 4 of those 50; the parse itself agrees on all 50. Strict metric: a transcript line starting with `[User]: ` followed by a host envelope marker.

- 122 of 1788 transcripts (6.8%) leak one: 103 AGENTS.md, 19 `<environment_context>`. Exactly one envelope line per transcript. 113 of the 122 are single-turn rollouts.

The AGENTS.md case, 103 of the 122, fails both checks in :107-:113, and the terminator at :113 is the one that has to move — accepting a bare header alone closes none of them, measured below. The message is one composite (redacted; this is also the fixture text used further down):

```
# AGENTS.md instructions

<INSTRUCTIONS>
# Project rules

- redacted
</INSTRUCTIONS>
<environment_context>
  <cwd>/repo</cwd>
  <shell>bash</shell>
  <current_date>2026-09-18</current_date>
  <timezone>Europe/Berlin</timezone>
  <filesystem><workspace_roots><root>/repo</root></workspace_roots><permission_profile>redacted</permission_profile></filesystem>
</environment_context>
```

`body.startswith("<INSTRUCTIONS>")` holds for all 103; `body.endswith("</INSTRUCTIONS>")` holds for none, because the text continues into an `<environment_context>` block. Corpus-wide, 666 of the 669 AGENTS.md-headed user messages end with `</environment_context>` and contain `<cwd>`; only 3 end at `</INSTRUCTIONS>`.

The header is separately bare here: all 669 start with `# AGENTS.md instructions` and none carries the ` for <path>` suffix that :107 and :111 want. Four redacted fixtures, identical except for the envelope text, isolate both checks:

| header | envelope body ends with | result |
|---|---|---|
| `# AGENTS.md instructions for /repo` | `</INSTRUCTIONS>` | filtered |
| `# AGENTS.md instructions for /repo` | `</environment_context>` | emitted as `[User]` |
| `# AGENTS.md instructions` | `</INSTRUCTIONS>` | emitted as `[User]` |
| `# AGENTS.md instructions` | `</environment_context>` | emitted as `[User]` |

The last row is the shape my rollouts contain, and it fails both checks. Output for it:

```
=== Transcript of a conversation between User and Codex CLI ===
[User]: # AGENTS.md instructions

<INSTRUCTIONS>
# Project rules

- redacted
</INSTRUCTIONS>
<environment_context>
  <cwd>/repo</cwd>
  <shell>bash</shell>
  <current_date>2026-09-18</current_date>
  <timezone>Europe/Berlin</timezone>
  <filesystem><workspace_roots><root>/repo</root></workspace_roots><permission_profile>redacted</permission_profile></filesystem>
</environment_context>
[User]: What is 3+3?
[Codex]: 6
```

and for the first row, the shape the fixtures pin:

```
=== Transcript of a conversation between User and Codex CLI ===
[User]: What is 3+3?
[Codex]: 6
```

These fixtures are minimal, not replicas: they use the leading records of the most common real sequence (`task_started` -> `response_item[developer]` -> `response_item[user]` envelope), which 93 of the 122 begin with. The real question never follows the envelope directly in any of those 93 — a `world_state` record sits between them in 92, a second developer message in 1 — and real turns also carry `turn_context`, `token_usage_record` and `compacted` records. 15 of the 122 never set `saw_leading_developer` at all, 13 of those because the turn carries no developer record anywhere.

The standalone `<environment_context>` case, 19 of the 122: there are 387 such messages in the corpus, 5 carry `<cwd>`/`<shell>`, and the other 382 are `<current_date>` + `<timezone>` + `<filesystem><workspace_roots>`, so the four-field requirement at :118-119 fails for them.

What this costs to cover, measured over the same 1788 rollouts:

- accepting a bare `# AGENTS.md instructions` header and dropping the four-field requirement at :118-119: 122 -> 121. Not enough on its own, which is why I am reporting the terminator rather than just the header.
- also letting the AGENTS.md body end at `</INSTRUCTIONS>` **or** `</environment_context>`: 122 -> 18, with all 103 AGENTS.md cases closed. `pytest tests/test_codex_parse_rollout.py` stays at 41 passed with that change applied.

The residual 18 are all `<environment_context>` and look intentional rather than broken: 13 have no later user message in the turn, which `has_later_user_message` deliberately preserves (`test_parse_rollout_preserves_first_user_literal_host_envelope` pins that), and the other 5 have one but no leading developer record, so `saw_leading_developer` at :237 skips the filter.

Baseline for all of the above: worktree at fc0ef77, `pytest tests/test_codex_parse_rollout.py` 41 passed, python 3.14 locally (CI pins 3.10-3.13). So this reads as a fixture gap rather than a logic one: the fixtures pin a ` for <path>` header with an `<INSTRUCTIONS>`-terminated body, and a standalone `<environment_context>` carrying all four required fields. The rollouts here carry a bare header with a composite body, and 382 of their 387 standalone `<environment_context>` messages have no `<cwd>` or `<shell>`.

One code property I did **not** observe in this corpus, in case it matters when you scope the fix: the filter is guarded by `not saw_conversational_user` (:236, set at :248), so once one envelope is emitted a following one is emitted too, even in the shape the matcher recognizes. A fixture reproduces it — an unrecognized envelope followed by a ` for <path>` one emits both, while two recognized ones in a row are both filtered — but all 122 leaking transcripts have exactly one envelope line, so it compounds nothing here.

I can send the change as a PR against main with fixtures built from the observed shapes (redacted), or open it as an issue first if you would rather scope it yourself.

zc277584121:
Thanks for documenting the post-close host-envelope gap. We independently reproduced it and addressed it in https://github.com/zilliztech/memsearch/pull/754.

Issue #306, opened 2026-04-08
Milvus crash on hybrid_search: Assert "std::isfinite(element.val)" failed → "Invalid sparse row: NaN or Inf value"


qianshengzj:
### Description

When performing a search

### Steps to reproduce

memsearch search 111123

### Error output / logs

```shell
Assert "std::isfinite(element.val)"  => Invalid sparse row: NaN or Inf value at /workspace/milvus-lite/thirdparty/milvus/internal/core/src/common/Utils.h:269

2026-04-08 18:07:36,250 [ERROR][_log_rpc_error]: RPC error: [hybrid_search], <MilvusException: (code=2000, message=Assert "std::isfinite(element.val)"  => Invalid sparse row: NaN or Inf value at /workspace/milvus-lite/thirdparty/milvus/internal/core/src/common/Utils.h:269
: segcore error)>, <elapsed:15.0ms>
Traceback:
Traceback (most recent call last):
  File "/home/qdd/miniconda3/lib/python3.13/site-packages/pymilvus/decorators.py", line 404, in handler
    return func(*args, **kwargs)
  File "/home/qdd/miniconda3/lib/python3.13/site-packages/pymilvus/decorators.py", line 451, in handler
    return func(self, *args, **kwargs)
  File "/home/qdd/miniconda3/lib/python3.13/site-packages/pymilvus/decorators.py", line 342, in handler
    raise e from e
  File "/home/qdd/miniconda3/lib/python3.13/site-packages/pymilvus/decorators.py", line 305, in handler
    return func(*args, **kwargs)
  File "/home/qdd/miniconda3/lib/python3.13/site-packages/pymilvus/client/grpc_handler.py", line 1331, in hybrid_search
    return self._execute_hybrid_search(
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        hybrid_search_request, timeout, round_decimal=round_decimal, context=context, **kwargs
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/qdd/miniconda3/lib/python3.13/site-packages/pymilvus/client/grpc_handler.py", line 1187, in _execute_hybrid_search
    raise e from e
[...]

PR #745, opened 2026-09-13
fix: count stored rows, not sealed row_count, in search and stats (#534)

### LINKED ISSUE #534: MilvusStore.upsert() reports success but writes are not durable on remote Milvus 2.5+ (missing flush)
## Summary

`MilvusStore.upsert()` returns `len(chunks)` as a fallback success count even when the underlying upsert produced no durable rows in Milvus. The `memsearch` CLI then reports e.g. `Indexed 219 chunks` while `get_collection_stats()` shows `row_count: 0`. All subsequent searches return empty results because nothing was actually queryable.

The issue surfaces specifically against **remote Milvus 2.5+ standalone**. Milvus-lite (the embedded path) appears to auto-flush, masking the bug.

## Environment

- memsearch: `0.4.2` (latest as of 2026-05-09, installed via `claude plugin install memsearch`)
- Milvus: `milvusdb/milvus:v2.5.4` standalone (Docker, embedded etcd)
- pymilvus: bundled with memsearch
- Client: Windows 11 (uvx installed) connecting to remote Milvus on Linux VM via `milvus.uri = "http://<vm>:19530"`
- Embeddings: `embedding.provider = "onnx"` (all-MiniLM-L6-v2, dim=384)

## Reproduction

1. Run a remote Milvus 2.5.x standalone:
   ```bash
   docker run -d --name milvus -p 19530:19530 -p 9091:9091 \
     -e ETCD_USE_EMBED=true -e ETCD_DATA_DIR=/var/lib/milvus/etcd \
     -e ETCD_CONFIG_PATH=/milvus/configs/embedEtcd.yaml \
     -e COMMON_STORAGETYPE=local \
     -v $PWD/milvus/volumes/milvus:/var/lib/milvus \
     -v $PWD/milvus/embedEtcd.yaml:/milvus/configs/embedEtcd.yaml \
     milvusdb/milvus:v2.5.4 milvus run standalone
   ```

2. Configure memsearch (`~/.memsearch/config.toml`):
   ```toml
   [milvus]
   uri = "http://<vm-ip>:19530"

   [embedding]
   provider = "onnx"
   ```

3. Index a project:
   ```bash
   memsearch index <project-path>
   # Output: Indexed 219 chunks across 47 files (or similar)
   ```

4. Inspect Milvus:
   ```python
   from pymilvus import MilvusClient
   c = MilvusClient(uri='http://<vm-ip>:19530')
   print(c.get_collection_stats('memsearch_chunks'))
   # Actual:   {'row_count': 0}
   # Expected: {'row_count': 219}
   ```

5. Search returns nothing:
   ```bash
   memsearch search "anything"
   # No matches found.
   ```

## Evidence the bug is in memsearch, not Milvus or pymilvus

I ran a direct minimal pymilvus test against the same Milvus instance, with the same schema memsearch uses (BM25 sparse function + 384-dim FLOAT_VECTOR + COSINE + SPARSE_INVERTED_INDEX):

```python
from pymilvus import MilvusClient, DataType, Function, FunctionType
client = MilvusClient(uri='http://<vm>:19530')
# (schema setup — same shape as memsearch._ensure_collection)
client.upsert(collection_name=COL, data=[{
    'chunk_hash': 'test1',
    'embedding': [0.1] * 384,
    'content': 'hello world',
    'file_path': 'test.md',
    'chunk_index': 0,
    'tags': '',
    'metadata': '{}'
}])
client.flush(COL)
print(client.get_collection_stats(COL))
```

Result: `{'row_count': 1}`. **Direct path works.** The minimal change vs. memsearch is the explicit `flush()` call.

## Root cause

`memsearch/store.py:125-137`:

```python
def upsert(self, chunks):
    if not chunks:
        return 0
    result = self._client.upsert(collection_name=self._collection, data=chunks)
    return result.get(\"upsert_count\", len(chunks)) if isinstance(result, dict) else len(chunks)
```

Two issues:

1. **No `flush()` after upsert.** Milvus 2.5+ does not auto-flush in the standalone path. Data sits in the in-memory write buffer and is invisible to queries until a flush, segment seal, or compaction. Older Milvus versions (and milvus-lite) auto-flushed for small writes — masking this gap.

2. **Silent fallback to `len(chunks)`.** If Milvus returns a dict without `upsert_count` (e.g. on the 2.5+ schema-mismatch error path some configs hit), the function still returns the *requested* count as success. `core.py:194` (`return self._store.upsert(records)`) then reports it back to the user. There's no integrity check that `row_count` increased.

## Suggested fix

```python
def upsert(self, chunks):
    if not chunks:
        return 0
    result = self._client.upsert(collection_name=self._collection, data=chunks)
    self._client.flush(self._collection)  # Required for Milvus 2.5+ standalone

    # Trust upsert_count over len(chunks) — fail loud if the server says 0
    actual = result.get(\"upsert_count\") if isinstance(result, dict) else None
    if actual is None:
        actual = len(chunks)
    if actual == 0 and len(chunks) > 0:
        raise RuntimeError(f\"Upsert reported 0 writes for {len(chunks)} chunks (silent failure)\")
    return actual
```

Optional follow-up: a periodic batch flush (every N chunks or on `index` completion) instead of per-call, to amortize the cost on large indexing runs.

## Impact

This is silent data loss from the user's perspective — the CLI reports success, the `~/.memsearch/<project>/chunks/` markdown layer is intact, but the vector index is empty so all searches return nothing. Diagnosis required reading `pymilvus` server logs to confirm zero `Insert`/`Upsert` RPCs were ever executed, despite memsearch's claim of 219 successful writes.

## Workaround for users

Until fixed, run `client.flush('memsearch_chunks')` manually after every `memsearch index` call.

---

Filed by a Claude Code user debugging memsearch + remote Milvus integration. Happy to test a patch if you'd like.

DrTrippel:
Fixes #534

## Problem
On Milvus Server, `memsearch search` and `memsearch stats` report a collection as empty for about 10 minutes after `memsearch index` succeeds. The data is not lost. memsearch just doesn't see it.

`MilvusStore.search()` guards against the #306 BM25 crash by returning `[]` when `get_collection_stats()["row_count"] == 0`, and `count()` returns the same value. `row_count` only counts **sealed/flushed** segments. Freshly upserted rows sit in a growing segment until the DataNode syncs them (`dataNode.segment.syncPeriod`, 600s by default). Until then:
- `row_count` is 0, so search returns nothing without querying;
- `stats` prints `Total indexed chunks: 0`.

Collections that already have an older flushed segment are not affected: the guard passes, and `hybrid_search` sees the growing rows immediately. That is why the bug only shows up on new or freshly re-created collections.

Milvus Lite flushes eagerly, which masks the problem, so #534 looked like a missing flush.

## Repro (Milvus v3.0.1 standalone in Docker, memsearch 0.4.20, pymilvus 3.0.1)
| Step | Result |
|---|---|
| `memsearch index <dir> --collection ms_probe` | `Indexed 1 chunks.` |
| `get_collection_stats("ms_probe")["row_count"]` | `0` |
| `query("ms_probe", filter="", output_fields=["count(*)"], consistency_level="Strong")` | `1` |
| `memsearch search <canary> --collection ms_probe` | `No results found.` until **608s** after indexing, then found |
| Existing 23-row collection, `stats` / `search` | `0` / nothing; after `MilvusClient.flush(...)`, `23` and a score-1.0 hit |

## Fix
- `count()` counts stored rows with `query(filter="", output_fields=["count(*)"], consistency_level="Strong")`.
- `search()` guards on `self.count() == 0`, so the #306 protection for truly empty collections stays. It also stops counting deleted-but-not-compacted rows.

## Tests
**New tests**
- `test_unsealed_rows_are_counted_and_searchable` simulates Milvus Server by patching `get_collection_stats` to report `row_count: 0` after an upsert. It fails on `main` (`assert 0 == 1`) and passes with this change.
- `test_search_empty_collection_returns_empty` keeps the #306 guard covered.

**Local runs**
- `uv run python -m pytest tests/test_store.py`: 19 passed.
- `uv run python -m pytest`: 423 passed, 7 skipped, 6 failed. The same 6 fail on unmodified `main` (`test_claude_stop_indexes_only_server[*]`, `test_session_start_and_data_commands_are_scoped_to_host_project[*]`), so they are unrelated to this change.
- `ruff check` and `ruff format --check`: clean.

https://claude.ai/code/session_01Ng6tyJJrBWCrSZNKGUbBdc


src/memsearch/store.py
```diff
@@ -190,8 +190,7 @@ def search(
         from pymilvus import AnnSearchRequest, RRFRanker
 
         # BM25 crashes on empty collections (avgdl=0 → NaN). See #306.
-        stats = self._client.get_collection_stats(self._collection)
-        if int(stats.get("row_count", 0)) == 0:
+        if self.count() == 0:
```
yhmo, line 193: src/memsearch/store.py line:193
Medium ---- search() now calls self.count() on every invocation, and the new count() issues a query(..., output_fields=["count(*)"], consistency_level="Strong") aggregation RPC. This replaces a cheap get_collection_stats() metadata call with a collection-wide count(*) aggregation plus an extra round-trip on the hot search path (CLI search, plugin memory-recall hooks, Memory.search). For the emptiness guard, row_count > 0 already proves the collection is non-empty (this is exactly the case that worked before #534), so the aggregation query only needs to run when row_count == 0 to distinguish a truly empty collection from one whose rows are still unsealed/growing. Consider a fast path that returns [] when get_collection_stats()["row_count"] > 0, and only falls back to self.count() when row_count == 0.

tests/test_store.py
```diff
@@ -49,6 +49,36 @@ def test_upsert_and_search(store: MilvusStore):
...
+                "source": "fresh.md",
+                "heading": "",
+                "chunk_hash": "fresh1",
+                "heading_level": 0,
+                "start_line": 1,
+                "end_line": 1,
+            }
+        ]
+    )
+    monkeypatch.setattr(store._client, "get_collection_stats", lambda *a, **k: {"row_count": 0})
```
yhmo, line 70: tests/test_store.py line:70
Low ---- The docstring says this test "simulates Milvus Server" by patching get_collection_stats to report row_count: 0, but the new count() never calls get_collection_stats. On Milvus Lite both counters agree after an upsert, so this test only proves that count()/search() no longer depend on row_count (a valid regression guard); it cannot detect a regression where query(..., output_fields=["count(*)"]) stops seeing unsealed/growing rows on a real Milvus Server. The server-side behavior at the core of #534 is currently only covered by the manual repro in the PR body. Consider clarifying the test scope in the docstring or adding a Server-backed integration test.

zc277584121:
Thanks for investigating this and for putting together the initial fix. We’ve merged #757, which keeps the metadata fast path, only falls back to a Strong count when stats report zero, adds malformed-count handling and Milvus Server coverage, and documents the remaining mixed-collection consistency window. I’m closing this PR as superseded by #757.

